PR #151, opened 2015-01-19
Support for verifiers (currently PEVerify)

GeertvanHorrik:


GeertvanHorrik:
Copying all refs works. I know it's not the best solution, but it works...

SimonCropp:
yeah but i am concerned about unintended side effects of other tools that make assumptions about the state of the obj directory

GeertvanHorrik:
I understand, just making babysteps to check where we can improve.

SimonCropp:
so there is not a safe subset of known error codes about missing refs we can pass into the `/ignore` param of peverfiy?

If not i think we should go with copy the files to a staging dir for verification, or plug into the after build. Thoughts?

GeertvanHorrik:
Yes, about missing refs there is. But there are more. For example:

```
*.Controls.AdvertisementControl::.cctor][offset 0x0000000A] Unable to resolve token.
```

You don't want to miss these. In this case it's a WIN 8.1 project with shared projects (shouldn't make a difference though). When I copy references, all works.

SimonCropp:
`Unable to resolve token` ok agreed that could be a valid error

GeertvanHorrik:
1>      Verifying assembly
1>      Finished verification in 525ms.

Not bad for an assembly with 700 types and 70 references (don't ask, it now copies all System.* assemblies as well).

SimonCropp:
add a new verification task and add it to the afterbuild similar to what is done here https://github.com/Fody/Fody/blob/master/NuGet/Fody.targets#L45

GeertvanHorrik:
ok, will work on that, thanks for the hints.

SimonCropp:
no worries. the code should be pretty simple in the end. 

* task takes in the `$(TargetPath)` in msbuild
* resolves peverify
* log errors

Since u dont need to worry about appdomain juggling like it need to do for the weavers

SimonCropp:
well also do config switching :)

GeertvanHorrik:
Introducing Configuration class so we can share configuration over multiple tasks.

GeertvanHorrik:
Also need to add FinalAssemblyFileName to support the final output path.

GeertvanHorrik:
1>      Verifying assembly
1>      Finished verification in 459ms.

woohoo! :-)

GeertvanHorrik:
btw. I prepared the Configuration object so it accepts DefinedConstants. Then we can customize the configuration based on both xml **and** constants in a single place for both tasks.

GeertvanHorrik:
Did you find time to review it?

SimonCropp:
will try to get to it over the weekend

SimonCropp:
manually merged this and deplyed a beta.

I called yagni on multiple verifiers. when we think of a second one we can plug that in

GeertvanHorrik:
1>      Verifying assembly
1>      Finished verification in 335ms.

Works great :-)

SimonCropp:
I will test more and move it out of beta today

PR #191, opened 2015-07-27
Unity (.NET 3.5) Compatability

Kleptine:
Right now only FodyIsolated and FodyCommon compile under .NET 3.5. I could probably get the rest running with a little more time, but it was unnecessary for my purposes (only using InnerWeaver). 

Most of these changes are accommodating lacks of functionality by either adding missing features, or rewriting to avoid compile issues (lambda/action/func took a little bit of finagling to get it to both compile and not crash on runtime).

It works for me in Unity -- let me know if you have any desire to start to merge this into the main Fody. I'd be willing to look into getting the last project compatible if you're interested. I'll admit it does make things a little bit messy.


SimonCropp:
sorry but i dont think it is the correct approach to be asking a project that already supports 5+ years of .net versions to support further back. Any effort should be put into getting unity to support .net 4

PR #197, opened 2015-08-18
Fix weaver project search path in Linux

ylatuya:
Create the search path in a portable way, since a path with '\' is not valid in Linux and Directory.Exists will always return False.

### DIFF
--- a/Fody/WeaverProjectFileFinder.cs
+++ b/Fody/WeaverProjectFileFinder.cs
@@ -24,7 +24,8 @@ public partial class Processor
 
     void GetValue()
     {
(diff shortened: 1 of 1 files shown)


SimonCropp:
deployed to nuget. 

Good catch BTW

SimonCropp:
https://www.nuget.org/packages/Fody/1.29.3

ylatuya:
Awesome! Thanks for the fast response :D

SimonCropp:
fast? it took me 7 hours. fast in open source is measured in minutes ;)

PR #214, opened 2015-12-09
Fixed setting inherited props on weaver

jasonholloway:
Hi there,

I'm working on an intermediate add-in to supply further (vaguely planned) add-ins with functionality, and decided the best way to do this would be by ModuleWeaver base class.

I've not completely proved the idea to myself yet, but this was the first hurdle: a slight bug in building the property-setting delegates, which meant that accessors for inherited properties weren't being compiled.

Thanks!

### DIFF
--- a/FodyIsolated.Tests/DelegateBuilders/PropertyDelegateBuilderTests.cs
+++ b/FodyIsolated.Tests/DelegateBuilders/PropertyDelegateBuilderTests.cs
@@ -2,7 +2,7 @@ using System;
 using NUnit.Framework;
 
 [TestFixture]
-public class PropertyDelegateBuilderTests
+public class PropertyDelegateBuilderTests : BaseClassSupplyingInheritedMembers
 {
     public string Property { get; set; }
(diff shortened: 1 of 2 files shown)


SimonCropp:
wouldnt it be better to provide these features by composition rather than inheritance?

jasonholloway:
Well, yes, you have a point... but the way ModuleWeaver is bound by convention, composition would require the user to make loads of explicit requests for resources, which would then have to be individually fed into whatever component required them. This is hardly impossible, I know, but I'm eager to suppress Cecil and things like assembly resolution as much as possible, leaving the user with a nice clean surface for doing what they want without all the lower-level bits intruding.

What'd be perfect (in, like, an ideal world or something) would be something a bit like OWIN's AppBuilder, where everything's available in a well-defined type, and can be neatly accessed and manipulated by extensions. That is, just enough strong typing to make composition nice. I think this is what I'm really trying to achieve with the populating of inherited properties.

PR #218, opened 2016-01-03
add nuget NUnitTestAdapter to Test Projects so VS Test Eplorer works

0x53A:
This allows running the Unit Tests in the VS TestAdapter:

image

### DIFF
--- a/FodyCommon.Tests/FodyCommon.Tests.csproj
+++ b/FodyCommon.Tests/FodyCommon.Tests.csproj
@@ -31,9 +31,25 @@
     <WarningLevel>4</WarningLevel>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="nunit.core, Version=2.6.4.14350, Culture=neutral, PublicKeyToken=96d09a1eb7f44a77, processorArchitecture=MSIL">
+      <HintPath>..\packages\NUnitTestAdapter.2.0.0\lib\nunit.core.dll</HintPath>
+      <Private>False</Private>
+    </Reference>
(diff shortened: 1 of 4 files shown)


SimonCropp:
R u serious that it is 2016 and VS test still doesn't pick up the most popular testing lib?

0x53A:
Not inbox. They built some kind of framework where ut-frameworks can register their own adapters. There are also adapter-nugets for xunit, etc.

Just out of curiosity, how do you run the unit tests?

SimonCropp:
in the past i have used http://www.testdriven.net/ but now i use R#.

But i see this is also popular https://visualstudiogallery.msdn.microsoft.com/6ab922d0-21c0-4f06-ab5f-4ecd1fe7175d

On the CI server i use either AppVeyor or TeamCity. 

Sorry but i dont think i can accept this PR. it should not be up to every open source project out there to add comparability hacks for test runners that refuse to innovate.

0x53A:
Well, they did kind of innovate, before VS2013 the test-runner was hardcoded for MSTest only, now they have an adapter framework for other systems.

The advantage of the nuget adapter is that it just works out of the box with a fresh VS install, no extensions required.

But I can totally understand that you don't accept this, I don't really care either way, since I'm not (yet =) ) a regular contributor, I just thought it might make the unit tests more discoverable.

SimonCropp:
> The advantage of the nuget adapter is that it just works out of the box with a fresh VS install, no extensions required.

The could also make this work OOTB in VS and remove all the friction.

PR #228, opened 2016-03-11
ability to select a custom fody xml file

negue:
I'm using costura and I was in need of having two fody xml files. 

For example in `debug` build I need a library to debug, but in release I don't really need this library / dll in my combined exe.

Example to have multiple xml's in csproj:

```
  <Target Name="MyFodyOverrideTarget">
    <Message Text="MyFodyOverrideTarget execute" />
    <Message Text="Old File Path $(CustomFodyFile)" />
    <CreateProperty Value="$(ProjectDir)FodyWeavers.$(Configuration).xml">
      <Output TaskParameter="Value" PropertyName="CustomFodyFile" />
    </CreateProperty>
     <CreateProperty Value="$(ProjectDir)">
      <Output TaskParameter="Value" PropertyName="FodyPath" />
    </CreateProperty>
    <Message Text="New File Path $(CustomFodyFile)" />
  </Target>
  <PropertyGroup>
    <FodyDependsOnTargets>MyFodyOverrideTarget</FodyDependsOnTargets>
  </PropertyGroup>
```

### DIFF
--- a/Fody/Processor.cs
+++ b/Fody/Processor.cs
@@ -14,6 +14,7 @@ public partial class Processor
     public string ProjectDirectory;
     public string References;
     public string SolutionDirectory;
+    public string CustomFodyFile;
     public string NuGetPackageRoot;
     public List<string> ReferenceCopyLocalPaths;
     public List<string> DefineConstants;
@@ -64,6 +65,11 @@ public partial class Processor
         ValidateAssemblyPath();
 
         ConfigFiles = ConfigFileFinder.FindWeaverConfigs(SolutionDirectory, ProjectDirectory, Logger);
+
+      if (!string.IsNullOrEmpty(CustomFodyFile))
+      {
+         ConfigFiles.Add(CustomFodyFile);
+      }
 
         if (!ShouldStartSinceFileChanged())
         {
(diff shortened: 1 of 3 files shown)


SimonCropp:
Merging a lib with costura does not stop u from debugging into it

negue:
Yes thats correct.

But the library is use is about 1MB big, which I don't need in my release :wink:

SimonCropp:
> But the library I use is about 1MB big, which I don't need in my release

sorry i dont understand this statement

Enumix:
Hello @SimonCropp , I'm using Stamp.Fody in solutions with large number of projects. It's very useful for specifying exactly which version running on production, but it takes some time. I would like to stamp only projects in Release mode. I've been trying to find the way to have different weavers configuration for Debug and Release mode. Sadly there is no way to embed FodyWeavers into app.config (I would prefer approach with config templates). 

Is there any workaround to do that? If now, why you not accepting this pull request?

kamil-mrzyglod:
@Enumix - are you using any kind of build server? An easy workaround is having a separate `FodyWeavers` file, which is being used when building a release build of your application(in fact it's something I've been using for a while).

SimonCropp:
@Enumix i suggest you give https://github.com/GitTools/GitVersion a try instead

PR #235, opened 2016-04-10
Padded addded to the list of addins

Scooletz:


Scooletz:
That was quick. Thx! :+1:

PR #239, opened 2016-05-04
Add a link to With plugin

mikhailshilkov:
With.Fody addin extends immutable C# classes with `With()` methods to return a copy of an object with one property modified.


SimonCropp:
thanks. can u also send a PR here with a sample project https://github.com/Fody/FodyAddinSamples/

PR #262, opened 2016-09-17
Renamed example powershell scripts to have the ps1 extension

Keboo:
When using the scripts in WeaverNugetTemplates in a NuGet packages they need to be named install.ps1 and uninstall.ps1 for NuGet to pick them up.


Keboo:
Any feedback on this?

PR #263, opened 2016-09-20
Support running Fody with Code Contracts

### LINKED ISSUE #259: Running Fody before Code Contracts
I'm trying out using Fody together with Code Contracts (via NuGet).

To get them playing nicely, I wanted to ensure that Fody ran before Code Contracts (so that weavers like NullGuard had a chance to do their thing).

There was a couple of things I needed to do to make this work:

Ensure that Code Contracts was imported in the .csproj file after Fody

```
  <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
  <PropertyGroup>
    <CodeContractsInstallDir>$([System.IO.Path]::GetFullPath('$(MSBuildThisFileDirectory)..\..\packages\DotNet.Contracts.1.10.20606.1\'))</CodeContractsInstallDir>
  </PropertyGroup>
```

then

```
  <Import Project="..\..\packages\Fody.1.29.4\build\dotnet\Fody.targets" Condition="Exists('..\..\packages\Fody.1.29.4\build\dotnet\Fody.targets')" />
  <Import Project="$(CodeContractsInstallDir)\MsBuild\v$(VisualStudioVersion)\Microsoft.CodeContracts.targets"/>

```

Change Fody's .targets file to hook into CompileDependsOn instead of AfterTarget.

```
  <PropertyGroup>
    <CompileDependsOn>$(CompileDependsOn); FodyTarget</CompileDependsOn>
  </PropertyGroup>
```

If that looks reasonable, I'm happy to look at creating a PR. If there's a better way to do that, I'm happy to learn!

flcdrg:
If the Code Contracts extension or NuGet package is used in a project, it does an additional compilation stage and produces a secondary assembly for analysis. Fody needs to run against this extra assembly otherwise the Code Contracts analysis stage will flag warnings for things that are potentially resolved by a Fody addin in the primary output assembly.

Fixes #259

### DIFF
--- a/NuGet/Fody.targets
+++ b/NuGet/Fody.targets
@@ -75,6 +75,28 @@
       
   </Target>
 
+  <Target
+      Name="FodyTargetForCodeContracts"
+      AfterTargets="ContractDeclarativeAssembly">
+
+      <Fody.WeavingTask
+          AssemblyPath="@(ContractDeclarativeAssembly->'%(FullPath)')"
+          IntermediateDir="$(IntermediateDir)"
+          KeyFilePath="$(FodyKeyFilePath)"
+          NuGetPackageRoot="$(NuGetPackageRoot)"
+          ProjectDirectory="$(ProjectDir)"
+          SolutionDir="$(FodySolutionDir)"
+          References="@(ReferencePath)"
+          SignAssembly="$(FodySignAssembly)"
+          ReferenceCopyLocalPaths="@(ReferenceCopyLocalPaths)"
+          DefineConstants="$(DefineConstants)"
+      >
+
+      <Output
+        TaskParameter="ExecutedWeavers"
+        PropertyName="FodyExecutedWeavers" />
+    </Fody.WeavingTask>
+  </Target>
 
   <UsingTask
     TaskName="Fody.VerifyTask"


SimonCropp:
wont this make the weaving task run twice?

flcdrg:
Yes, that's correct - as there is the secondary assembly that needs to be processed.

kamil-mrzyglod:
Can't you import such custom `.targets` file into your `.csproj` and make it run `AfterTargets="ContractDeclarativeAssembly"`?

flcdrg:
@kamil-mrzyglod Sure - that's what I'm doing at the moment, but that won't help anyone else who wanted to use Code Contracts together with Fody.

kamil-mrzyglod:
If it makes Fody runs twice, even if this is not needed, than it's a bit overkill IMO ;)

flcdrg:
Yes, it would be great to run it once, but as far as I can tell, that's not how Code Contracts works - it generates a second assembly and in order for the code contract analysis stage to work, Fody needs to have an opportunity to modify that second assembly first.

Note that this change should only affect projects that have Code Contracts enabled. Any other project would not have the Code Contracts target so it would be ignored.

SimonCropp:
sorry this is bringing too much knowledge of codecontracts into this assembly. suggest you raise this with codecontracts  team and see if they can come up with a better solution

flcdrg:
I'm not sure why adding one extra MSBuild target is "too much knowledge" as opposed to just "knowledge" - especially since the target is ignored except when Code Contracts are being used.

But as it's apparently not a common problem I'll just stick with my own local change. If there's suddenly huge interest in using Fody and Code Contracts in the future, then the fix will be here if someone needs it 😄

PR #264, opened 2016-09-27
Don't attempt to weave for intellisense builds

kzu:
Otherwise, this breaks Roslyn when loading the project, which causes further issues 
in intellisense and build.


kzu:
Not convinced this solves the underlying issues, and it causes other problems. Discarding.

PR #299, opened 2017-03-29
Fix broken generated weaver result

kswoll:
Somewhat recently, Fody has started generating a `FodyWeaverResults` file that doesn't specify a base type which breaks `Assembly.GetTypes()` at runtime.  I tried creating a unit test, but the existing unit test `WeavingInfoTests.WeavedAssembly_ShouldContainWeavedInfo` currently fails, so not sure how to get at the fody-generated assembly for testing purposes.


kswoll:
The check is failing because master is currently in a broken state.

kswoll:
By the way, the error from PEVerify.exe is:

> [MD]: Error: TypeDef that is not an Interface and not the Object class extends Nil token.

My PR gets rid of it.

SimonCropp:
merged manually and included in 2.0

Issue #305, opened 2017-04-24
Adding Fody 2.0.2 to Xamarin.Forms app breaks the build with Microsoft.Build.BuildEngine.InvalidProjectFileException


renzska:
I created a clean Multi Platform Xamarin.Forms app. Built in debug mode and it worked fine. Added Fody 2.0.2 and built again and it errors with the following:

```
Error: Error building target AfterCompile: Microsoft.Build.BuildEngine.InvalidProjectFileException: Unable to parse condition "Exists(@(IntermediateAssembly))" : Expected an item list at position 8 in condition "Exists(@(IntermediateAssembly))". Missing opening parantheses after the '@'. ---> Microsoft.Build.BuildEngine.ExpressionParseException: Expected an item list at position 8 in condition "Exists(@(IntermediateAssembly))". Missing opening parantheses after the '@'.
  at Microsoft.Build.BuildEngine.ConditionParser.IsAtToken (Microsoft.Build.BuildEngine.TokenType type, System.String error_msg) [0x00077] in /private/tmp/source-mono-2017-02/bockbuild-2017-02/profiles/mono-mac-xamarin/build-root/mono-x86/mcs/class/Microsoft.Build.Engine/Microsoft.Build.BuildEngine/ConditionParser.cs:290 
  at Microsoft.Build.BuildEngine.ConditionParser.ParseReferenceExpression (System.Char prefix) [0x00023] in /private/tmp/source-mono-2017-02/bockbuild-2017-02/profiles/mono-mac-xamarin/build-root/mono-x86/mcs/class/Microsoft.Build.Engine/Microsoft.Build.BuildEngine/ConditionParser.cs:231 
  at Microsoft.Build.BuildEngine.ConditionParser.ParseFactorExpression () [0x000a5] in /private/tmp/source-mono-2017-02/bockbuild-2017-02/profiles/mono-mac-xamarin/build-root/mono-x86/mcs/class/Microsoft.Build.Engine/Microsoft.Build.BuildEngine/ConditionParser.cs:185 
  at Microsoft.Build.BuildEngine.ConditionParser.ParseFunctionArguments () [0x0005c] in /private/tmp/source-mono-2017-02/bockbuild-2017-02/profiles/mono-mac-xamarin/build-root/mono-x86/mcs/class/Microsoft.Build.Engine/Microsoft.Build.BuildEngine/ConditionParser.cs:219 
[...]

PR #306, opened 2017-04-24
Fix Microsoft.Build.BuildEngine.InvalidProjectFileException: Unable t…

renzska:
…o parse condition "Exists(@(IntermediateAssembly))" bug.

It seems that @Exists breaks when the variable inside may contain a period or space in the path. Adding single quotes resolves this issue.
https://github.com/Fody/Fody/issues/305


SimonCropp:
Thakns

SimonCropp:
this is now released

PR #348, opened 2017-05-19
Changed Nuget weaver package search to be case insensitive

### LINKED ISSUE #331: Weavers not found due to case sensitivity, nuget restore
Here is our scenario:

- We are building a mobile application via VSTS on a mac
- We are using Fody, and the PropertyChanged weaver
- We found that Fody fails to find the weaver in a lower case package name scenario, outlined below

build output for the Fody MSBuild target:
```
Target FodyTarget:
	Fody: Fody (version 1.29.4.0) Executing
	Fody: ProjectDirectory: '/buildagent/_work/2/s/App/DeveloperModule/'.
	Fody: AssemblyPath: '/buildagent/_work/2/s/App/DeveloperModule/obj/Release/DeveloperModule.dll'
	Fody: Found path to weavers file '/buildagent/_work/2/s/App/DeveloperModule/FodyWeavers.xml'.
	Fody: SolutionDirectory path is '/buildagent/_work/2/s/App/'
	Fody: Finding weavers
	Fody: Adding weaver dlls from '/buildagent/_work/2/s/App/Packages'.
	Fody: Could not find packages dir from nuget config.
	Fody: SolutionDirectoryPath: /buildagent/_work/2/s/App/
	Fody: Adding weaver dlls from '/Users/buildagent/.nuget/packages/fody'.
	Fody: SolutionDirectoryPath: /buildagent/_work/2/s/App/
	Fody: Skipped scanning '/buildagent/_work/2/s/App/Tools' for weavers since it doesn't exist.
	Fody: No Weaver project file found.
: error : Fody: Could not find a weaver named 'PropertyChanged'. 
```

After some troubleshooting, we found the following path was problematic: `/Users/buildagent/.nuget/packages/propertychanged.fody`

This is the standard nuget package cache location for the build agent user.

When we renamed the package, using:
`mv propertychanged.fody PropertyChanged.Fody` the build began working.

Unfortunately, the lower case behavior appears to be base behavior in Nuget 3.5.0.1938, which is what our VSTS Step runs, for example:

```
2017-05-10T18:09:18.6094670Z ##[section]Starting: NuGet restore developer module
2017-05-10T18:09:18.6263120Z ==============================================================================
2017-05-10T18:09:18.6275690Z Task         : NuGet Installer
2017-05-10T18:09:18.6288140Z Description  : Installs or restores missing NuGet packages
2017-05-10T18:09:18.6300480Z Version      : 0.2.31
2017-05-10T18:09:18.6312850Z Author       : Microsoft Corporation
2017-05-10T18:09:18.6325370Z Help         : [More Information](https://go.microsoft.com/fwlink/?LinkID=613747)
2017-05-10T18:09:18.6337990Z ==============================================================================
2017-05-10T18:09:19.1140180Z Detected NuGet version 3.5.0.1938 / 3.5.0
```

And we are hesitant to explore odd work-arounds like attempting to rename the folder after the fact on a set of shared build machines.

I would expect that locating the PropertyChanged weaver should not be case sensitive, as the ID is not treated as case sensitive by Nuget.

We use Fody heavily, and appreciate this excellent project. Thanks in advance for any assistance with this issue!

ColinDabritzViewpoint:
Fixes #331 

The Nuget weaver search, ScanNuGetPackageRoot, now pulls all packages and DLLs regardless of casing, while still matching the convention.

Packages are matched as either case:
* `<PackageName>.Fody`
* `<packagename>.fody`

DLLs are matched as either case:
* `<packagename>.fody.dll`
* `<PackageName>.Fody.dll`

Some of the filter is done in code instead of the file system pattern request, but it should still generally filter well up front.

The `WithNugetPackageRoot` test was updated. The fake weaver folder for weaver 2 was changed to lower case, but the DLLs inside are upper case. A new approval test capture was required, as the changed case value was updated.

This addresses the specific issue raised with nuget on mac using Mono to execute a build, and should make the process a little more robust in terms of case matching.

### DIFF
--- a/Fody.Tests/NuGetPackageRootTest.WithNuGetPackageRoot.approved.txt
+++ b/Fody.Tests/NuGetPackageRootTest.WithNuGetPackageRoot.approved.txt
@@ -1,6 +1,6 @@
 ﻿[
   "\\Weaver1.Fody\\1.0.7\\Weaver1.Fody.dll",
   "\\Weaver1.Fody\\1.0.8\\Weaver1.Fody.dll",
-  "\\Weaver2.Fody\\2.0.7\\Weaver2.Fody.dll",
-  "\\Weaver2.Fody\\2.0.8\\Weaver2.Fody.dll"
+  "\\weaver2.fody\\2.0.7\\Weaver2.Fody.dll",
+  "\\weaver2.fody\\2.0.8\\Weaver2.Fody.dll"
 ]
--- a/Fody/AddinFinder/AddinSearchDirectories.cs
+++ b/Fody/AddinFinder/AddinSearchDirectories.cs
@@ -16,7 +16,7 @@ public partial class AddinFinder
         }
         else
         {
-            foreach (var directory in PackageDefinitions.Where(x => x.ToLowerInvariant().Contains(".fody")))
+            foreach (var directory in PackageDefinitions.Where(x => x.ToLowerInvariant().EndsWith(".fody")))
             {
                 AddFiles(Directory.EnumerateFiles(directory, "*.Fody.dll"));
             }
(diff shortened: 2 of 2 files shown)


ColinDabritzViewpoint:
Fody.Tests passed locally in windows.

ColinDabritzViewpoint:
A colleague pointed out that the existing pattern doesn't end it a star, so Contains is incorrect to maintain the existing behavior. We also decided that the behavior with existing PackageDefinitions should likely be the same. I updated the PR to use 'EndWith' in both scenarios, so only `<packagename>.fody` will match, and `<packagename>.fody.cody` will not.

Tests pass locally.

ColinDabritzViewpoint:
I updated to 2.0.10 on our project and built with lowercase packages present. Thanks for your help with this issue, and thanks for making Fody a reality!

PR #361, opened 2017-06-04
Better symbols detection

### LINKED ISSUE #360: Fody >= 2.0.8 producing assemblies without debug information
While testing to see if #357 successfully enables the sourcelink scenario, I noticed it still was wasn't working. I tested the PDB, and it had all of the correct info. However, I noticed in the VS Modules debugger windows that it was showing that "Binary was not built with debug information", so it wasn't even trying to load the PDB.

After testing all of the recent releases to see where this started happening, 2.0.7 and earlier are producing assemblies that can load symbols correctly, but 2.0.8 and newer are all broken.

The older 2.0.x releases aren't tagged, but looking at the milestones it appears that #334 was the only change between 2.0.7 and 2.0.8. I verified that $(DebugSymbols) was true in the build I was using to test this, so it would appear that the newly added DebugSymbols property is false even when it shouldn't be, causing Fody to incorrectly skip writing the PDB.

bording:
The `DebugSymbols` MSBuild property can be false or not set even when debug symbols are still being produced, so using the `_DebugSymbolsProduced` property should be used instead to correctly detect when symbols have been created.

Fixes #360

### DIFF
--- a/FodyIsolated/SymbolsFinder.cs
+++ b/FodyIsolated/SymbolsFinder.cs
@@ -16,7 +16,7 @@ public partial class InnerWeaver
     {
         if (!DebugSymbols)
         {
-            // Log something about building without symbols?
+            Logger.LogWarning($"Debug symbols are disabled. It is recommended to build with debug symbols enabled.");
             return;
         }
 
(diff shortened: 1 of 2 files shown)


bording:
Closing this since another approach was taken!

PR #398, opened 2017-08-24
Documentation file path

### LINKED ISSUE #397: Pass path of XML-Documentation file to weaver
Weavers adding new types or members that are public visible should be able to extend the XML documentation to include the new content.

tom-englert:
Fixes #397

### DIFF
--- a/Fody/WeavingTask.cs
+++ b/Fody/WeavingTask.cs
@@ -5,6 +5,7 @@ using Microsoft.Build.Utilities;
 namespace Fody
 {
     using System;
+    using System.Windows.Forms;
 
     public class WeavingTask : Task, ICancelableTask
     {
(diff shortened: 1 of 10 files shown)


Fody/WeavingTask.cs
```diff
@@ -22,6 +22,8 @@ public class WeavingTask : Task, ICancelableTask
         [Required]
         public string ProjectDirectory { get; set; }
 
+        public string[] DocumentationFilePaths { get; set; }
```
SimonCropp: why is this an array?

tom-englert: Just tried, works with a string as well, since there is never more than one instance of `DocFileItem`. I'll update the PR.

Issue #416, opened 2017-10-24
Attribute to choose the weaver location


Logerfo:
According to [this]( https://github.com/Fody/Fody/wiki/InSolutionWeaving), and confirmed here: https://github.com/Fody/Fody/blob/3705c852155aa90013c2251f783c5860be1a9bef/Fody/WeaverProjectFileFinder.cs#L31
Fody looks for a reference named "Weavers" to find them, which causes me the following problem:
```
SolutionA
 -ProjectA
 -Weavers
SolutionB
 -ProjectB
```
ProjectA references Weavers.
ProjectB references ProjectA. Consequently, Projec B references Weavers, but won't be able to find it unless I reference it again, which causes a warning, but compiles.
Proposal: create an attribute to choose the weaver location. Example:
```xml
<Weavers>
   <ExampleWeaver Location="ProjectA" />
</Weavers>
```

Logerfo:
In my case, it's actually here:
https://github.com/Fody/Fody/blob/3705c852155aa90013c2251f783c5860be1a9bef/Fody/WeaverProjectFileFinder.cs#L40
But that's not the only place to change if this is happening. There's a lot of places, actually.

Logerfo:
PR sent https://github.com/Fody/Fody/pull/419

Issue #415, opened 2017-10-23
Fody fails to find assembly with the same name


Logerfo:
I have the current setup:
```
Solution1
  -Library1
  -Weavers1
Solution2
  -Library2
  -Weavers2
```
>Weavers1 depends on Library1.
Weavers2 depends on Library1 and Weavers1.
Library2 depends on Library1, Weavers2 and Weavers1. 

In my case, both weaver projects were named "Weavers", which caused the following error trying to build Solution2: `Could not load type 'Weavers.Utils' from assembly 'Weavers...`
Renaming any of the Weaver projects fixed the issue.

Logerfo:
Searching through the code I've found out the cause of this issue:
https://github.com/Fody/Fody/blob/3705c852155aa90013c2251f783c5860be1a9bef/Fody/WeaverProjectFileFinder.cs#L40
Fody only looks for the first reference with the name "Weavers.dll"
A solution for this may come by itself with the implementation requested in my other issue https://github.com/Fody/Fody/issues/416.

SimonCropp:
closed since https://github.com/Fody/Fody/pull/419#issuecomment-352729393

PR #419, opened 2017-10-25
Support for different weaver source

Logerfo:
See #416

This is a per-weaver configuration.
Before looking for in-solution or "Weavers.dll", Fody will look for an attribute named "Source" in the xml weaver element.

Usage: `<ExampleWeaver Source="ExampleSource"/>`
For this very ExampleWeaver, Fody will look for the first reference of ExampleSource.dll
if there are multiple assemblies named ExampleSource, Fody will only search for the weaver in the first it finds. This means this feature is also affected by #415.

### DIFF
--- a/Fody/InstanceLinker/WeaversConfiguredInstanceLinker.cs
+++ b/Fody/InstanceLinker/WeaversConfiguredInstanceLinker.cs
@@ -1,4 +1,7 @@
 
+using System;
+using System.Linq;
+
 public partial class Processor
 {
     public void ConfigureWhenWeaversFound()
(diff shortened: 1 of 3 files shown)


SimonCropp:
can u add some tests

Logerfo:
@SimonCropp I'm afraid not, since I have zero experience with unit testing

SimonCropp:
would it be possible for you to request some help from friend or colleagues to help you write tests?

SimonCropp:
ok while i want to accept this PR, i dont actually need this feature and dont have the time to finish this PR by writing tests, so how should we proceed?

Logerfo:
I already found my way without this, so you may close if you want

SimonCropp:
ok cool. thanks for getting back to me

PR #459, opened 2018-01-06
Fix #454

### LINKED ISSUE #454: Allow weavers to prevent writting assemblies.
Currently Fody [always writes](https://github.com/Fody/Fody/blob/master/FodyIsolated/InnerWeaver.cs#L103) back the module. 
For one of my weavers, I would like it if it were possible for the weaver to communicate back to Fody that writing the assembly should not be done (or is not needed). Because my weaver does not need the assembly written, I believe this would allow it to be used, without breaking the edit-and-continue in Visual Studio (am I wrong is this assumption?)

I am happy to submit a PR for this, but wanted to discuss how to best implement it (since there may be more than one weaver being applied). As present it does not appear that there is any precident for weavers communicating back information. Thoughts? Feedback? Suggestions?

Keboo:
Fixes #454 

Adding a new property that weavers can set to indicate that they did not modify the assembly. If all weavers set this to true, the assembly will not be re-written.

### DIFF
--- a/FodyHelpers/BaseModuleWeaver.cs
+++ b/FodyHelpers/BaseModuleWeaver.cs
@@ -110,6 +110,12 @@ namespace Fody
         public List<string> DefineConstants { get; set; }
 
         /// <summary>
+        /// Set to true if this weaver did not modifiy the ModuleDefinition.
+        /// If all weavers set this property to true, Fody will leave the assembly untouched.
+        /// </summary>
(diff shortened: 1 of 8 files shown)


SimonCropp:
sorry for the delay. can u rebase ?

Keboo:
@SimonCropp yes, I can do that soon

SimonCropp:
will close for now. happy to re-consider after a rebase

PR #461, opened 2018-01-07
Filter weavers by version number

Keboo:
Adding the ability to filter which version of a weaver gets loaded by specifying a "WeaverVersion" attribute as part of the weaver's element in the FodyWeavers.xml. Versions numbers can be specified as individual version numbers (including wild cards; "1.2.3.*"), or as a range ("[1.2.3, 2.*)").

This is technically a breaking change for anyone who was previously using this attribute for their own purpose (unlikely).

I find the current behavior especially problematic on repositories where the packages directory is not checked in. When switching between branches where different version of the weaver are expected to be used, Fody always grabs the latest version which forces me to clear previous versions of the weaver just to build my projects correctly. This would allow specifying the appropriate version in the FodyWeavers.xml so that the code in each branch of my project would use the appropriate weaver version.

### DIFF
--- a/Fody/AddinFinder/AddinFilesEnumerator.cs
+++ b/Fody/AddinFinder/AddinFilesEnumerator.cs
@@ -2,16 +2,20 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using Fody;
 
 public partial class AddinFinder
 {
     public List<string> FodyFiles = new List<string>();
 
-    public string FindAddinAssembly(string packageName)
+    public string FindAddinAssembly(string packageName, VersionFilter versionFilter)
     {
         var packageFileName = packageName + ".Fody.dll";
-        return FodyFiles.Where(x => string.Equals(Path.GetFileName(x), packageFileName, StringComparison.OrdinalIgnoreCase))
-            .OrderByDescending(AssemblyVersionReader.GetAssemblyVersion)
-            .FirstOrDefault();
+        return (from file in FodyFiles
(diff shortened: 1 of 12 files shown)


SimonCropp:
would it be possible for you to instead update to the new sdk style of csproj?

Keboo:
So only in some cases is switching an option. I do a lot of WPF work which forces the old csproj.

However, I am not sure how switching to the new csproj helps (perhaps I am missing something obvious?).
Unless I am not understanding the code correctly, the [AddinFinder](https://github.com/Fody/Fody/blob/master/Fody/AddinFinder/AddinSearchDirectories.cs) simply parses through the various possible directories and [selects the latest](https://github.com/Fody/Fody/blob/master/Fody/AddinFinder/AddinFilesEnumerator.cs#L14) version of any given weaver. The issue I am facing is, at times, I do not want the latest version of my weaver used. Perhaps I am working on fixing a bug on a production branch of code so I want it to use the same weaver version that was used originally (even though there may be multiple versions in my nuget directory). 
Is there some way to select a particular weaver version using the new csproj?

SimonCropp:
@Keboo 

> However, I am not sure how switching to the new csproj helps (perhaps I am missing something obvious?).

the new sdk csproj passes the nuget versions through to the msbuild task. hence the correct nuget is always used

SimonCropp:
Unfortunately i cant accept this pr. It leaks in too many implementation details of nuget versioning. This has casue significant pain and bugs in the past.

I will however accept a pr that leverages the version of the refence assembly (which is already passed in to fody) to infer the weaver version.

PR #485, opened 2018-02-21
Fix NCrunch support

ltrzesniewski:
Looks like NCrunch support was broken by the `netstandard`/`netclassic` directory split. This fixes the issue by including the correct directory instead of the (now basically empty) root one.


SimonCropp:
Good catch. Thanks

SimonCropp:
this is now deployed

PR #490, opened 2018-03-06
ConfigureAwait package is no longer maintained

MarcBruins:
It's currently unlisted on nuget(https://www.nuget.org/packages/ConfigureAwait.Fody)


SimonCropp:
i will release a v2 of ConfigureAwait shortly

Issue #491, opened 2018-03-08
Fody InSolutionWeaving does not work with ProduceReferenceAssembly


0x53A:
With ``<ProduceReferenceAssembly>true</ProduceReferenceAssembly>`` (see https://github.com/dotnet/roslyn/blob/master/docs/features/refout.md), Fody fails with

```
2>------ Rebuild All started: Project: UseTheWeaver, Configuration: Debug Any CPU ------
2>    Fody: Fody (version 2.4.4.0) Executing
2>MSBUILD : error : Fody: An unhandled exception occurred:
2>MSBUILD : error : Exception:
2>MSBUILD : error : Cannot load a reference assembly for execution.
2>MSBUILD : error : Type:
2>MSBUILD : error : System.BadImageFormatException
2>MSBUILD : error : StackTrace:
2>MSBUILD : error :    at System.Reflection.RuntimeAssembly.nLoadImage(Byte[] rawAssembly, Byte[] rawSymbolStore, Evidence evidence, StackCrawlMark& stackMark, Boolean fIntrospection, SecurityContextSource securityContextSource)
2>MSBUILD : error :    at System.Reflection.Assembly.Load(Byte[] rawAssembly)
2>MSBUILD : error :    at InnerWeaver.LoadFromFile(String assemblyPath)
2>MSBUILD : error :    at InnerWeaver.LoadAssembly(String assemblyPath)
2>MSBUILD : error :    at InnerWeaver.InitialiseWeavers()
2>MSBUILD : error :    at InnerWeaver.Execute()
2>MSBUILD : error : Source:
2>MSBUILD : error : mscorlib
2>MSBUILD : error : TargetSite:
2>MSBUILD : error : System.Reflection.RuntimeAssembly nLoadImage(Byte[], Byte[], System.Security.Policy.Evidence, System.Threading.StackCrawlMark ByRef, Boolean, System.Security.SecurityContextSource)
2>MSBUILD : error : 
2>    Fody:   Finished Fody 116ms.
```

### Proposed fix:

ReferenceAssemblies have a special Attribute: ``System.Runtime.CompilerServices.ReferenceAssemblyAttribute``.

Extend the dll loader logic to skip all assemblies with that attribute.

-----------------

Complete as many of the following as possible

 * [x] Uploaded a minimal solution that reproduces the problem with the fewest moving pieces. Preferably a public GitHub repository: https://github.com/0x53A/FodyCustomWeaverAndRefAssemblies
[...]

PR #492, opened 2018-03-08
[Ready For Review] #491: ignore ref-assemblies when searching for InSolution weaver dll.

0x53A:
fixes https://github.com/Fody/Fody/issues/491

-------

The unit test failed because [Weavers.dll](https://github.com/Fody/Fody/blob/master/Tests/Fody/WeaversProjectFileFinder/WithWeaver/Weavers/bin/Debug/Weavers.dll) was actually a html file...


SimonCropp:
can u add a test for this?

0x53A:
The easiest test would be to add ``<ProduceReferenceAssembly>true</ProduceReferenceAssembly>`` to [Weavers.csproj](https://github.com/Fody/Fody/blob/master/Integration/Weavers/Weavers.csproj).

Actually we would then only test the ref-assembly case, but otherwise I would need to add an additional solution.

What do you think?

0x53A:
@SimonCropp please see https://github.com/Fody/Fody/pull/493, but TL;DR, I don't know how to actually add a test.

PR #523, opened 2018-04-28
Add support for delay-signing with a public only keyfile

ryknuth:
Problem:
Delay-signing allows for keyfiles to contain only the public key since the private key is not needed. Unfortunately, public key bytes do not create a valid StrongNameKeyPair.  StrongNameKeyPair requires the private key to regenerate the public key. Mono.Cecil uses the StrongNameKeyPair to set the public key on the assembly and to sign the assembly.  If an invalid StrongNameKeyPair is given to Mono.Cecil, an exception is thrown and Fody fails.

Fix:
Avoid creating the StrongNameKeyPair and instead use the new PublicKeyBytes on the WriterParameters from Mono.Cecil. If provided, Mono.Cecil will directly set these bytes on the AssemblyName and avoid attempting to sign the assembly.

It's a bit difficult to tell what type of keyfile is present. The simplest way is to create the StrongNameKeyPair and attempt to generate the public key. If this throws an ArgumentException, we know that the private key was not present within the keyfile.  In this case, null out the StrongNameKeyPair and pass along the key file bytes as the public key bytes.


ryknuth:
The API depends on Mono.Cecil accepting my PR for them.

SimonCropp:
Pushing to your branch should automatically update this pr. U should see the commit in the commits tab of the pr

ryknuth:
I would love to add tests.  Could you give me a good location in your test suites to add some?

SimonCropp:
@ryknuth unfortunately i cant accept this until it is the associated cecil PR is merged into the cecil master

SimonCropp:
i have updated cecil to 33cfa97. can u rebase this

Kethku:
This is currently blocking me. Is there anyway I can help to push this along?

SimonCropp:
@Kethku i merged this. but it is blocked (ie i cant release) until this PR in cecil is merged https://github.com/jbevain/cecil/pull/513

Issue #532, opened 2018-07-28
Weaver using symbols can't be tested using WeaverTestHelper.ExecuteTestRun


tom-englert:
WeaverTestHelper.ExecuteTestRun does not provide symbols to the weaver under test because the ReaderParameters are not setup properly and symbols are not written back.

PR & Tests will follow.

tom-englert:
@ltrzesniewski doesn't it? Maybe the changes above already fix it?

ltrzesniewski:
I admit I haven't tested this yet, but there's no reference to `WriterParameters.WriteSymbols` in `WeaverTestHelper`, so I assumed it wouldn't work.

tom-englert:
I see, but maybe you'd better create a new issue, I would not like to commit that under the wrong title.

ltrzesniewski:
I'd say it's related, since "using symbols" could be understood as "reading/writing symbols".
But if you'd rather not include the writing part here, I'll submit a PR when I'll have some time (I'm writing this from a tablet right now...).

tom-englert:
OK, maybe your'e right, I'll add that to the same PR.

ltrzesniewski:
Thanks! :smile:

SimonCropp:
@tom-englert thanks for the PR. this is now deployed

ltrzesniewski:
FYI this change uncovered an interesting problem in my weaver. It should be pretty uncommon but I thought I'd mention it here just in case.

Turns out writing debug information can influence the metadata Cecil emits.

InlineIL is commonly used by declaring `using static InlineIL.IL.Emit;`, and the `using static` directive differs from a simple `using` in that it *will* reference the metadata token of the imported type, not simply its name: the `ImportScope` table in the pdb will contain a `TypeRef` token, so when writing symbols, Cecil needs to add the required row to the `TypeRef` table. Other *"unusual"* directives such as `extern alias` seem to have a similar behavior.

If you then remove the `AssemblyRef` to the weaver library, for instance by setting `BaseModuleWeaver.ShouldCleanReference` to true, you'll end up with a `TypeRef` that references an invalid `AssemblyRef`. This makes PEVerify unhappy.

Here's what I use instead of `ShouldCleanReference` to work around the issue:

https://github.com/ltrzesniewski/InlineIL.Fody/blob/3bf0afe8dc7a53e4f1058af01e19529676165926/src/InlineIL.Fody/ModuleWeaver.cs#L87-L123

[...]

PR #533, opened 2018-07-28
Fix Issue#532

tom-englert:
Fix Issue #532


tom-englert:
@SimonCropp looks like you did not update the Cecil binaries...

PR #534, opened 2018-07-29
Import helpers

tom-englert:
Add extension methods to ease importing of system types in the weaver.


tom-englert:
Moved to https://github.com/tom-englert/FodyTools since this is more flexible

PR #538, opened 2018-08-16
Support delay signing through setting public key on ModuleDefinition

ryknuth:
We don't actually need to change cecil to support delay signing.  We can simply set the public key on

`ModuleDefinition.Assembly.Name`

Verified this works on a local project.


SimonCropp:
Is it possible to add some tests around this?

ryknuth:
Yes I would like to.  Could you direct me to where might be a good test to copy?  I was thinking about doing this just how I manually tested.  Run through the entire thing: use a fake weaver and then have Fody output the DLL.  Is there a test that does that?

ryknuth:
AppVeyor can't sign it looks like. Any thoughts?

IlyaFinkelshteyn:
Check https://github.com/ryknuth/Fody/pull/3. Not sure I fully understand why `dotnet build` does not like it, but this should unblock you.

SimonCropp:
seems cecil doesnst support strong naming on netcore, which is the runtime dotnet.exe runs in  https://github.com/jbevain/cecil/blob/eea822cad4b6f320c9e1da642fcbc0c129b00a6e/Mono.Cecil/ModuleDefinition.cs#L215

ryknuth:
Does this look good?  Can we get it merged into main?

SimonCropp:
@ryknuth @IlyaFinkelshteyn thanks. this is now deployed

PR #546, opened 2018-09-09
Read/write embedded pdb

### LINKED ISSUE #544: Embedded debug info is not supported
Fody assumes in many places that debug info *must* be in a separate file. If you process an assembly with embedded debug info, the debug info gets stripped away.

This means you can't really use both Fody and [SourceLink](https://github.com/dotnet/sourcelink) on the same assembly, as SourceLink doesn't seem to work reliably with an external portable PDB file (VS can't always find the PDB for some reason).

I thought about changing [the `DebugSymbols` property](https://github.com/Fody/Fody/blob/67be8226acc30c2c31761d009beb4cd038ba0440/Fody/WeavingTask.cs#L77) to an enum with values `None`/`Embedded`/`External`, and then using `debugReaderProvider = new EmbeddedPortablePdbReaderProvider()` when it's `Embedded` (same for the writer).

Does this approach seem OK?

ltrzesniewski:
Fixes #544.

Unfortunately, this will fail with the current Cecil version. Cecil first needs to be updated with the following PR for the integration tests to pass: https://github.com/jbevain/cecil/pull/534

How would you like to proceed, given that Fody uses a "private" Cecil build?

### DIFF
--- a/Fody/Fody.targets
+++ b/Fody/Fody.targets
@@ -80,7 +80,6 @@
           ReferenceCopyLocalPaths="@(ReferenceCopyLocalPaths)"
           DefineConstants="$(DefineConstants)"
           PackageDefinitions="@(FodyPackageDefinitions)"
-          DebugSymbols="$(DebugSymbols)"
           DebugType="$(DebugType)"
           DocumentationFilePath="@(DocFileItem->'%(FullPath)')"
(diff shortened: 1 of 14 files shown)


ltrzesniewski:
The PR just got merged.

SimonCropp:
@ltrzesniewski thanks. i will try to get to that tonight

PR #552, opened 2018-09-22
Update InlineIL description

ltrzesniewski:
This changes the description of InlineIL.

I removed the reference to `ILGenerator` since the InlineIL API evolved into something completely different a while ago, and no longer tries to mimic `ILGenerator`. The previous description seemed outdated.


SimonCropp:
thanks :)

PR #557, opened 2018-09-27
Support probing path

tom-englert:
This is the initial version supporting the `WeaverProbingPaths` idea discussed on gitter.

It is done with minimal changes to the existing code and fully backward compatible.
It currently improves selecting the correct weaver if the weaver provides the `.props` file and so adds itself to the `WeaverProbingPaths` poperty.
When there are enough weavers supporting this, I will start optimizing the code, to look at the probing paths first and omit the other search paths if all weavers have been located.

It can be tested with https://www.nuget.org/packages/JetBrainsAnnotations.Fody/, versions 2.3 and 2.4 have added support for this.


SimonCropp:
can the approach u used in JetBrainsAnnotations.Fody be pushed to other weavers via FodyPackaging https://github.com/Fody/Fody/tree/master/FodyPackaging

Fody/AddinFinder/AddinFilesEnumerator.cs
```diff
@@ -7,11 +7,24 @@ public partial class AddinFinder
...
+
     public string FindAddinAssembly(string packageName)
     {
         var packageFileName = packageName + ".Fody.dll";
-        return FodyFiles.Where(x => string.Equals(Path.GetFileName(x), packageFileName, StringComparison.OrdinalIgnoreCase))
-            .OrderByDescending(AssemblyVersionReader.GetAssemblyVersion)
+
+        var candidates = FodyFiles.Where(x => string.Equals(Path.GetFileName(x), packageFileName, StringComparison.OrdinalIgnoreCase))
+            .OrderBy(ProbingPathScore)
+            .ThenByDescending(VersionReader);
```
SimonCropp: rather than re parsing the version on every package lookup. shouldnt this be parsed and stored once?

tom-englert: It was like this before, so I assumed this will be called only once per weaver.

tom-englert:
Yes, I already thought about extending FodyPackaging, just had no time to look at it yet...

Fody/AddinFinder/AddinFilesEnumerator.cs
tom-englert: This is just step 1: make it work
I'm still working on optimizing this.

tom-englert:
OK, I think that's it.

tom-englert:
I'll be off for a week, then I can update readme and docu

LG
Tom
________________________________
Von: Simon Cropp<mailto:notifications@github.com>
Gesendet: ‎30.‎09.‎2018 06:27
An: Fody/Fody<mailto:Fody@noreply.github.com>
Cc: tom-englert<mailto:mail@tom-englert.de>; Author<mailto:author@noreply.github.com>
Betreff: Re: [Fody/Fody] Support probing path (#557)


Merged #557<https://github.com/Fody/Fody/pull/557> into master.

Issue #560, opened 2018-10-01
NuGet install of Fody 3.2.6 on VS2015 corrupts project/fails to install


syserr0r:
#### Describe the bug

Versions 3.2.5 and 3.2.6 of Fody do not install using Nuget on VS2015 when installed as an update or a dependency of another package -- the error is `The default XML namespace of the project must be the MSBuild XML namespace. If the project is authored in the MSBuild 2003 format, please add xmlns="http://schemas.microsoft.com/developer/msbuild/2003" to the <Project> element. If the project has been authored in the old 1.0 or 1.2 format, please convert it to MSBuild 2003 format.  <project-path>\packages\Fody.3.2.6\build\Fody.targets`

If one attempts to install Fody (not as an update or dependency) then it may appear to succeed, but the project will become corrupted -- closing and re-opening the project will cause it to fail to load with the following error:

`The default XML namespace of the project must be the MSBuild XML namespace. If the project is authored in the MSBuild 2003 format, please add xmlns="http://schemas.microsoft.com/developer/msbuild/2003" to the <Project> element. If the project has been authored in the old 1.0 or 1.2 format, please convert it to MSBuild 2003 format.  <project-path>\packages\Fody.3.2.6\build\Fody.targets`

This also applies to Fody.PropertyChanged version 2.5.12 (with the error mentioning that package and .targets file instead)

#### Minimal Repro

1. Make new `WPF Application` project
2. Install Fody using NuGet package manager (which will attempt to install Fody3.2.6)
3. Close VS and attempt to re-open project (Project will fail to load)
(To allow the project to open again, delete the `Fody.3.2.6` folder from `packages/`)

Alternatively:
1. Make new `WPF Application` project
2. Attempt to install PropertyChanged.Fody (which will attempt to install Fody3.2.6 and PropertyChanged.Fody.2.5.12)
3. Receive error and install is abandoned (project unaltered)

#### Additional context

Reverting to older versions of Fody and Fody.PropertyChanged appears to work-around this issue.

My understanding is that VS2015 should be supported.
I am currently using VS 2015 Version 14.0.25431.01 Update 3

SimonCropp:
Why cant u move to vs2017?

syserr0r:
I can, I haven't got around to it yet though -- if you're no longer supporting VS2015 it might be worth mentioning as much

SimonCropp:
The issue template does state that i should be on the current stable version on Visual Studio 

[...]

PR #561, opened 2018-10-03
Add NCrunch support to weavers by default

ltrzesniewski:
This makes the addins play nicely with NCrunch by default.

I also inverted netclassicweaver/netstandardweaver because I noticed my "fulll .NET" MSBuild was using the netstandardweaver version.

I renamed  `RuntimeToken` to `WeaverRuntimeToken` to make it more specific to Fody, just in case.

Finally, I couldn't reproduce any issues by eliminating the `FilesToAdd` intermediate item, so I removed the indirection.

### DIFF
--- a/Fody/Fody.targets
+++ b/Fody/Fody.targets
@@ -73,9 +73,10 @@
       />
   </Target>
 
-  <!--Support for ncrunch-->
+  <!-- Support for NCrunch -->
   <ItemGroup  Condition="'$(NCrunch)' == '1'">
     <None Include="$(FodyAssemblyFile)\*.*" />
+    <None Include="@(WeaverFiles)" />
   </ItemGroup>
 
 </Project>
--- a/FodyPackaging/Weaver.Props.Template
+++ b/FodyPackaging/Weaver.Props.Template
@@ -1,12 +1,17 @@
 ﻿<Project>
+
   <PropertyGroup>
-    <RuntimeToken Condition="$(MSBuildRuntimeType) == 'Core'">netclassicweaver</RuntimeToken>
-    <RuntimeToken Condition="$(MSBuildRuntimeType) != 'Core'">netstandardweaver</RuntimeToken>
+    <WeaverRuntimeToken Condition="$(MSBuildRuntimeType) != 'Core'">netclassicweaver</WeaverRuntimeToken>
+    <WeaverRuntimeToken Condition="$(MSBuildRuntimeType) == 'Core'">netstandardweaver</WeaverRuntimeToken>
   </PropertyGroup>
+
   <ItemGroup>
-    <FilesToAdd Include="$(MsBuildThisFileDirectory)..\$(RuntimeToken)\$(MSBuildThisFileName).dll" />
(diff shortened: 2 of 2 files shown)


SimonCropp:
Did u confirm that items are being passed in to the fody task, ie by looking at the build log in diognostic mode

ltrzesniewski:
I can see the following line in the diagnostic logs after I've modified the PropertyChanged.Fody.props file in my local nuget cache:

```
8>  Task Parameter:WeaverFiles=C:\Dev\dotnet\lib\packages\PropertyChanged.Fody.2.5.12\build\..\netstandardweaver\PropertyChanged.Fody.dll (TaskId:482)
```

This is using the old csproj format.

ltrzesniewski:
Wait, that `netstandardweaver` is not good. I'll take a look.

ltrzesniewski:
Nevermind, I blame caching for this, as I've edited the props file in the cache. It works fine after restarting VS.

```
1>  Task Parameter:WeaverFiles=C:\Dev\dotnet\lib\packages\PropertyChanged.Fody.2.5.12\build\..\netclassicweaver\PropertyChanged.Fody.dll (TaskId:41)
```

I've also checked the binary log from the build, which shows the correct path:
image

ltrzesniewski:
One last thing I'm not sure about: `MSBuildRuntimeType` can also have the `Mono` value. Should it use the classic or standard weaver in that case?

ltrzesniewski:
I'm really sorry for the spam, but it just occurred to me there may be a better approach for adding NCrunch support to weavers. I'll try it in a few hours, please don't merge yet.

SimonCropp:
lol. people contributing is not spam :)  will hold off

ltrzesniewski:
I was referring to that bunch of messages I posted that started to feel like a monologue 😉 

Anyway, I made the change I wanted: I moved the items needed for NCrunch to Fody.targets, because it feels much better that way - there's less logic in the weaver props.

I also added the xmlns to the weaver props because of #560.

Finally, I didn't do anything about the `'$(MSBuildRuntimeType)' == 'Mono'` case I asked about, as the logic used by the weavers is the same that Fody.targets uses which is what matters I think.

Unfortunately, you'll still have to update all addins because netclassicweaver and netstandardweaver were inverted. 😕

ltrzesniewski:
BTW there was a disadvantage of using `FilesToAdd`: when using multiple weavers, each one would add an item to `FilesToAdd` and then append all `FilesToAdd` items to `WeaverFiles`.

Here's an example with 3 weavers:

image

The task would get duplicate weaver inputs.

Getting rid of `FilesToAdd` fixes this issue:

image

SimonCropp:
thanks heaps for this

ltrzesniewski:
@SimonCropp I'm not sure if you missed this detail or if you didn't have the time to do it yet, but since you didn't acknowledge it, I'd like to remind you that *all* addins need to be updated to reference the latest Fody version, because the `netclassicweaver`/`netstandardweaver` directories were inverted.

I guess you must have a script for that 🙂

SimonCropp:
Done

ltrzesniewski:
Thanks! 👍

PR #569, opened 2018-10-14
Fix #568: Support updating of ReferenceCopyLocalPaths.

### LINKED ISSUE #568: Support updating of ReferenceCopyLocalPaths
Currently the weavers `ReferenceCopyLocalPaths` is only an in parameter; weavers that make changes to the reqiured references, like e.g. Costura, currently have to provide their own build task to update the `ReferenceCopyLocalPaths` MSBuild variable and remove the obsolete references - which is a huge overhead.

It would be much easier if Fody would provide the mechanism used in Costura by default, i.e. reading the `ReferenceCopyLocalPaths` back from every weaver, and at the end passing the modified list back to MSBuild.

This way a weaver could simply remove any obsolete item from the `ReferenceCopyLocalPaths` list, and at the end Fody would notify MSBuild of the changes.

@SimonCropp If you agree that this is a valuable feature, I will try to solve this and provide a PR.

tom-englert:


tom-englert:
As soon as you release this, I can provide a PR for Costura to make use of it.

SimonCropp:
There seem to be some unrelated assembly version changes in this pr?

tom-englert:
@SimonCropp can please post some details on what you meant by "There seem to be some unrelated assembly version changes in this pr?" so I can fix it?

Fody/AddinFinder/AddinSearchDirectories.cs
```diff
@@ -149,16 +148,19 @@ struct AssemblyAndVersion
...
     {
         foreach (var versionDirectory in DirectoryEx.EnumerateDirectoriesContains(directory, ".fody."))
         {
             var fileName = Path.GetFileName(versionDirectory);
             var index = fileName.IndexOf(".fody.", StringComparison.OrdinalIgnoreCase);
             var packageName = fileName.Substring(0, index + 5);
-            var version = Version.Parse(fileName.Substring(index + 6).Split('-')[0]);
 
-            yield return new AssemblyAndVersion{Version = version, Assembly= GetAssemblyFromNugetDir(versionDirectory, packageName)};
+            if (Version.TryParse(fileName.Substring(index + 6).Split('-')[0], out var version))
```
SimonCropp: i dont understand why this change is in this PR?

SimonCropp:
can u also rebase this

Fody/AddinFinder/AddinSearchDirectories.cs
tom-englert: I got a crash here because there was a file in the .nuget cache that did not have a valid version in it's name, so I simply replaced `Version.Parse` with `Version.TryParse` to skip files that do not match the convention instead of crashing Fody.

tom-englert: Just removed a blank line, so now the diff shown in GitHub shows better what has changed.

tom-englert: I have moved this change from this PR to #573

SimonCropp:
this should not be deployed. pending a nuget cache update

PR #573, opened 2018-10-19
Fix #572

### LINKED ISSUE #572: Crash when a weaver folder name does not comply to expected pattern.
Crashes here if some folder contains `.fody.` but does not continue with a valid version string:
https://github.com/Fody/Fody/blob/a850433071857c3dcc9342fa21664b8ec89c8106/Fody/AddinFinder/AddinSearchDirectories.cs#L159

tom-englert:
skip folders that contain the string .fody. but do not continue with a valid version string


SimonCropp:
this should not be deployed. pending a nuget cache update

PR #588, opened 2018-10-28
remove unused solution configurations

### LINKED ISSUE #587: Remove unused/unmaintained solution configurations
Simplify maintenance by removing unused configurations from the solution

tom-englert:
Fixes #587


tom-englert:
this was just on click to remove it - if it only would be always so easy :smirk:

PR #595, opened 2018-10-30
Do not require a solution for FodyPackaging to work

ltrzesniewski:
FodyPackaging misbehaves when `$(SolutionDir)` is not set (for instance when building a single project), because an empty path is passed to `[MSBuild]::GetDirectoryNameOfFileAbove`, which throws the following error:

> error MSB4184: The expression "[MSBuild]::GetDirectoryNameOfFileAbove('', .git\index)" cannot be evaluated. The path is not of a legal form. 

This PR fixes that by using `$(MSBuildProjectDirectory)` as a starting point instead.

I also noticed that `$(LocalGitRootFolder)` should have a trailing slash, so this change ensures that as well.


SimonCropp:
this should now be deployed

Issue #505, opened 2018-03-25
Race condition in FodyPackaging


ltrzesniewski:
Suppose you have an addin named `SomeAddin.Fody` which uses a library named `SomeAddin` (the usual Fody addin setup). There is no project dependency between `SomeAddin` and `SomeAddin.Fody`.

You add a reference to `FodyPackaging` to `SomeAddin`, and it will build a NuGet package which will include the output of `SomeAddin` and `SomeAddin.Fody`.

The issue is that if MSBuild builds these two projects *in parallel* to each other (due to a lack of dependency), there will be a race condition if the `SomeAddin` project finishes building before `SomeAddin.Fody` is done. The build process will try to package a file which has not been built yet, and it will produce a package with missing files.

This can be solved by adding a build dependency `SomeAddin` -> `SomeAddin.Fody`, but this is not documented anywhere AFAICT.

I can see there's already a workaround to something that looks like it's related to this, but it doesn't seem to solve this particular issue for me:

```XML
  <ItemGroup>
    <!-- this project targets netstandard2.0 (unnecessarily) and has the below ref to work around this bug https://github.com/Microsoft/msbuild/issues/2661 -->
    <ProjectReference Include="..\$(PackageId)\$(PackageId).csproj" PrivateAssets="All" Condition="$(TargetFramework)=='fake'" />
  </ItemGroup>
```

If I disable parallel builds, run a cleanup and *then* a rebuild, I end up with no weaver in the package, because `SomeAddin` builds *before* `SomeAddin.Fody`.

This does seem to do the trick for me though:

```XML
  <ItemGroup>
    <ProjectReference Include="..\$(PackageId)\$(PackageId).csproj" IncludeAssets="None">
      <ReferenceOutputAssembly>false</ReferenceOutputAssembly>
    </ProjectReference>
  </ItemGroup>
```

With parallel builds disabled, `SomeAddin.Fody` now builds before `SomeAddin`.

But I'm not quite sure what exactly you were trying to solve with that conditional reference, so I don't know if that's the right solution.

SimonCropp:
ReferenceOutputAssembly is the correct solution. but it will only work for VS. it is still broken in `dotnet.exe` and `msbuild.exe` https://github.com/dotnet/sdk/issues/939 so i cant fix properly until 2.1.3xx.

until then u can package using VS or work around your race condition by forcing the parallele build to 1.

ltrzesniewski:
[...]

PR #597, opened 2018-11-03
Fix #596: Packaging now fails if no weavers are included in package

### LINKED ISSUE #596: FodyPackaging generates package without weavers
Du to wrong build order FodyPackaging generated the package before the weaver was built.
It created an invalid package without weavers, but it should have failed.

tom-englert:
Fix #596: Packaging now fails if no weavers are included in package

### DIFF
--- a/FodyPackaging/build/FodyPackaging.targets
+++ b/FodyPackaging/build/FodyPackaging.targets
@@ -9,7 +9,12 @@
       <NetClassicFilesToInclude Include="$(WeaverDirPath)\net4*\$(PackageId).pdb" />
       <NetStandardFilesToInclude Include="$(WeaverDirPath)\netstandard2*\$(PackageId).dll" />
       <NetStandardFilesToInclude Include="$(WeaverDirPath)\netstandard2*\$(PackageId).pdb" />
+
+      <TfmSpecificPackageFile Include="$(WeaverPropsTempFile)" PackagePath="build" />
     </ItemGroup>
+
+    <Error Text="FodyPackaging: No NetClassic weavers found to include in package. Maybe the build order is wrong?" Condition="'@(NetClassicFilesToInclude)'==''" />
+    <Error Text="FodyPackaging: No NetStandard weavers found to include in package. Maybe the build order is wrong?" Condition="'@(NetStandardFilesToInclude)'==''" />
 
     <Copy SourceFiles="$(WeaverPropsTemplate)" DestinationFiles="$(WeaverPropsTempFile)" />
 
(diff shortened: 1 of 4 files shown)


FodyPackaging/build/FodyPackaging.targets
```diff
@@ -9,19 +9,19 @@
...
-      <Output TaskParameter="Include" ItemName="TfmSpecificPackageFile" />
-    </CreateItem>
-    <CreateItem Include="$(WeaverPropsTempFile)" AdditionalMetadata="PackagePath=build">
-      <Output TaskParameter="Include" ItemName="TfmSpecificPackageFile" />
-    </CreateItem>
+    <PropertyGroup>
+      <AllWeaverFilesIncluded>@(NetClassicFilesToInclude)@(NetStandardFilesToInclude)</AllWeaverFilesIncluded>
+    </PropertyGroup>
+
+    <Error Text="FodyPackaging: No weavers found to include in package. Maybe the build order is wrong?" Condition="'$(AllWeaverFilesIncluded)' == ''" />
```
ltrzesniewski: Wouldn't it be better to raise an error if any one of the weavers is missing? I mean when either netstandard or netclassic is missing.
Having one but not the other could happen if you build projects in parallel.

tom-englert: Good point, I'll gonna change this.

SimonCropp:
Thanks. This should now be deployed pending a nuget update

ltrzesniewski:
This can unfortunately cause false positives.

Here's what I was doing previously (simplified):

```
msbuild src\InlineIL.sln /t:Restore 
msbuild src\InlineIL.sln /t:Rebuild
msbuild src\InlineIL\InlineIL.csproj /t:Pack
```

So, I first builded everything (with `GeneratePackageOnBuild` set to `false`), then I explicitly created the NuGet package. This always produced valid packages.

This change triggered the new error on the `Rebuild` step. 😢 

My current workaround is this (again, simplified):

```
msbuild src\InlineIL.sln /t:Restore
msbuild src\InlineIL.Fody\InlineIL.Fody.csproj /t:Build
msbuild src\InlineIL.sln /t:Build
msbuild src\InlineIL\InlineIL.csproj /t:Pack
```

I need to explicitly build the weaver project first. I couldn't find any other way to make this work 😢 

 - Setting project dependencies on the solution works in VS but fails on the command line
 - Adding a project reference with `ReferenceOutputAssembly="false" SkipGetTargetFrameworkProperties="true" GlobalPropertiesToRemove="TargetFramework"` as explained in https://github.com/Microsoft/msbuild/issues/3843 causes the same errors as if I had set project dependencies in VS
 - Using `dotnet build` instead of `msbuild` doesn't change anything (understandably)

So I'm basically experiencing #505 once again here 😞 

I have no clue how you guys manage to get reliable builds in your weavers without doing any of this. I couldn't find any hidden tricks. Am I missing something obvious here?

I'll submit a PR to postpone the check until the actual packing takes place.

ltrzesniewski:
Disregard my comment above. Turns out I haven't noticed that `FodyPackaging` sets `GeneratePackageOnBuild` to true.

PR #598, opened 2018-11-04
Feature #577

### LINKED ISSUE #577: Running Fody without FodyWeavers.xml
This came to my mind after reading #571.
I already tried to start this as a discussion on gitter, but there was no response:

Now that Fody knows about most of the installed weaver packages due to #557 (except for some legacy ones, that did not update yet), we could simplify the usage regarding the FodyWeavers.xml file.

One could just install the NuGet package and compile the project, and for weavers that do not need any configuration it would work out of the box.

I could think of several options:

- Run without FodyWeavers.xml: run all well known installed weavers with an empty config
- Create a default FodyWeavers.xml containing all well known installed weavers if it does not exist
- Update an existing FodyWeavers.xml with missing well known installed weavers (would maybe require the possibility to disable a weaver)

Any thoughts about this?

tom-englert:
Fix #577: 
- If there is no FodyWeavers.xml, one is created and the build fails with a message to review the file
- A FodyWeavers.xsd is created/updated on every build.
- Every weaver can contribute to this schema, so you will be able to easily configure your weavers without navigating to the weavers homepage and reading the documentation.

Currently a weaver exposes it's schema fragment through an embedded resource with the name `FodyWeavers_SchemaFragment`. I'm not fully happy with this solution, but it would need some major refactoring to improve this...


Fody/ConfigFileFinder.cs
```diff
@@ -18,23 +24,117 @@ public static List<string> FindWeaverConfigs(string solutionDirectoryPath, strin
         var projectConfigFilePath = Path.Combine(projectDirectory, "FodyWeavers.xml");
         if (!File.Exists(projectConfigFilePath))
         {
-            logger.LogDebug($@"Could not file a FodyWeavers.xml at the project level ({projectConfigFilePath}). Some project types do not support using NuGet to add content files e.g. netstandard projects. In these cases it is necessary to manually add a FodyWeavers.xml to the project. Example content:
+            if (!files.Any() && (wellKnownWeaverFiles != null))
+            {
+                GenerateDefault(projectConfigFilePath, wellKnownWeaverFiles);
+
+                logger.LogError($@"Could not file a FodyWeavers.xml at the project level ({projectConfigFilePath}). A default file has been created. Please review the file and add it to your project.");
```
ltrzesniewski:
```suggestion
                logger.LogError($@"Could not find a FodyWeavers.xml file at the project level ({projectConfigFilePath}). A default file has been created. Please review the file and add it to your project.");
```

Fody/ConfigFileFinder.cs
```diff
@@ -18,23 +24,117 @@ public static List<string> FindWeaverConfigs(string solutionDirectoryPath, strin
...
+            }
+            else
+            {
+                logger.LogDebug($@"Could not file a FodyWeavers.xml at the project level ({projectConfigFilePath}). Some project types do not support using NuGet to add content files e.g. netstandard projects. In these cases it is necessary to manually add a FodyWeavers.xml to the project. Example content:
```
ltrzesniewski: Same as above, but we probably also should log this as an error.

```suggestion
                logger.LogError($@"Could not find a FodyWeavers.xml file at the project level ({projectConfigFilePath}). Some project types do not support using NuGet to add content files e.g. netstandard projects. In these cases it is necessary to manually add a FodyWeavers.xml to the project. Example content:
```

Fody/ConfigFileFinder.cs
```diff
@@ -18,23 +24,117 @@ public static List<string> FindWeaverConfigs(string solutionDirectoryPath, strin
...
+            return;
+
+        var weaverConfigTemplate = Fody.Properties.Resources.FodyWeavers_FileTemplate;
+
+        var weaverEntries = wellKnownWeaverFiles
+            .Select(WeaverNameFromFilePath)
+            .Aggregate(new StringBuilder(), (accumulate, value) => accumulate.AppendFormat("\r\n  <{0} />", value))
+            .ToString();
+
+        var contents = string.Format(CultureInfo.InvariantCulture, weaverConfigTemplate, weaverEntries);
```
ltrzesniewski: Wouldn't it be easier to just build an `XDocument` from scratch there?

tom-englert: Yes, good points, I have refactored this. 
Using just a file simplifies everything. I just put it side by side with the assembly, so it's the same location when testing a weaver.

Fody/ConfigFileFinder.cs
```diff
@@ -18,23 +25,152 @@ public static List<string> FindWeaverConfigs(string solutionDirectoryPath, strin
...
+
+        var fragments = wellKnownWeaverFiles.Select(CreateItemFragment);
+
+        baseNode.Add(fragments);
+
+        var filePath = Path.ChangeExtension(projectConfigFilePath, ".xsd");
+
+        try
+        {
+            if (File.Exists(filePath) && string.Equals(XDocument.Load(filePath).ToString(SaveOptions.OmitDuplicateNamespaces | SaveOptions.DisableFormatting), schema.ToString(SaveOptions.OmitDuplicateNamespaces | SaveOptions.DisableFormatting)))
```
ltrzesniewski: This test is quite expensive... OK, it shouldn't really be noticeable during a build, but is it really necessary? Why wouldn't we always overwrite the schema?

ltrzesniewski:
What would you think about adding something like that to `Fody.targets`? (untested)

```xml
<ItemGroup>
  <None Update="$(MSBuildProjectDirectory)\FodyWeavers.xsd">
    <Visible>false</Visible>
  </None>
</ItemGroup>
```

This would hide `FodyWeavers.xsd` from the solution explorer, as the user shouldn't edit it anyway. Do you think it would be a good idea?

tom-englert:
good idea!

Fody/ConfigFileFinder.cs
tom-englert: It looks like overhead at the first glance, but if you take into account that writing is many times slower than reading, and that the file rarely changes, omitting the write will improve the performance in most of the cases. 

Also if a file has not changed, its timestamp should not change either - some SCMs may still rely on timestamps or file changes, or even make files read-only unless checked out explicitly.

ltrzesniewski: :+1:

ltrzesniewski:
I was thinking about one more thing: The xsd file could be placed in the `obj` directory.

Pros:
 - A schema file would not magically appear in the user source code
 - It would not be checked in into version control (it's better since it's a generated file)

Cons:
 - The `noNamespaceSchemaLocation` attribute file would point to a file in the `obj` directory

Since the file would not be checked in, the solution would need to be built once in order for completion to work. I think we could mitigate that by hooking into the design-time build and generating it at that time. This would still require the NuGet packages to be restored to work.

What do you think?

tom-englert:
I rather prefer to check in such non-voliatile generated files, like e.g. the *.designer.* files, and have it work out of the box even if I don't have compiled yet.

SimonCropp:
@tom-englert @ltrzesniewski 

so i dont have the capacity to review this. I have several fractures in my spine and can only sit at a computer for a short period of time.

so if u tell me is ready to go, i can do a merge and a release.

just let me know

SimonCropp:
same goes for the related downstream PRs

ltrzesniewski:
@SimonCropp OK I'll test this more thoroughly and let you know (so fat I've only reviewed the code).
Get well soon!

ltrzesniewski:
Ok so I tested this and it works well overall. I found one issue and I also have one idea for improvement:

## Issue: Booleans

`xs:boolean` doesn't cut it, because it accepts `true` and `false` (which are fine) but also `0` and `1` which are not recognized by Fody.

image

I suggest we change `xs:boolean` to `bool` and add its definition to the schema so it can be reused by the individual weaver element definitions:

```xml
  <xs:simpleType name="bool">
    <xs:restriction base="xs:string">
      <xs:enumeration value="true" />
      <xs:enumeration value="false" />
    </xs:restriction>
  </xs:simpleType>
```

Downstream PRs would have to be updated to use this new type. I noticed PropertyChanged.Fody throws an ugly error (_An unhandled exception occurred_) when you set `InjectOnPropertyNameChanged="1"`, so let's not suggest the users to do that. 😉 

Also, `True` won't be considered as a valid value by IntelliSense, but I don't think it's a real issue.

## Design: xcf

Ok so this is not exactly a problem, more of a design decision. but I mention it here because once this PR is shipped it will have to stay that way. Also, sorry for not thinking about that earlier. 😕

The xcf files only contains a single `xs:complexType` root element (and VS complains that _"The required attribute 'name' is missing"_). I think it would be better if this file were a full xsd schema.

That would let you define named types which could then be reused in several places by the weaver element definition. You'd also get designer support in VS (ok the designer isn't exactly great, but the experience would still be better).

Here's how merging could work:
- The schema would be expected to declare a single `xs:element` _direct child_ of its `xs:schema` root element, which would be merged into the `xs:all` element in the template. The `name` attribute would be expected to match the weaver name. `minOccurs="0" maxOccurs="1"` would be added during the merge.
- Any additional _direct children_ of `xs:schema` of types `xs:simpleType`, `xs:complexType`, `xs:attributeGroup` and `xs:group` would be added to the `xs:schema` in `FodyWeavers.xsd`
- Let's allow for redefinition of the `bool` type but ignore it (otherwise the standalone schema wouldn't be valid).

An example:

```xml
<?xml version="1.0" encoding="utf-8"?>
<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema">
  <xs:element name="SomeWeaver">
    <xs:complexType>
      <xs:attribute name="SomeOption" type="SomeWeaver.SomeEnum" />
      <xs:attribute name="SomeOtherOption" type="SomeWeaver.SomeEnum" />
    </xs:complexType>
  </xs:element>
  <xs:simpleType name="SomeWeaver.SomeEnum">
    <xs:restriction base="xs:string">
      <xs:enumeration value="Foo" />
      <xs:enumeration value="Bar" />
    </xs:restriction>
  </xs:simpleType>
</xs:schema>
```

Would be merged like the following:

```xml
<?xml version="1.0" encoding="utf-8"?>
<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema">
  <xs:element name="Weavers">
    <xs:complexType>
      <xs:all>
        <xs:element name="SomeWeaver" minOccurs="0" maxOccurs="1">
          <xs:complexType>
            <xs:attribute name="SomeOption" type="SomeWeaver.SomeEnum" />
            <xs:attribute name="SomeOtherOption" type="SomeWeaver.SomeEnum" />
          </xs:complexType>
        </xs:element>
      </xs:all>
      <xs:attribute name="VerifyAssembly" type="bool" />
      <xs:attribute name="VerifyIgnoreCodes" type="xs:string" />
    </xs:complexType>
  </xs:element>
  <xs:simpleType name="bool">
    <xs:restriction base="xs:string">
      <xs:enumeration value="true" />
      <xs:enumeration value="false" />
    </xs:restriction>
  </xs:simpleType>
  <xs:simpleType name="SomeWeaver.SomeEnum">
    <xs:restriction base="xs:string">
      <xs:enumeration value="Foo" />
      <xs:enumeration value="Bar" />
    </xs:restriction>
  </xs:simpleType>
</xs:schema>
```

Which is basically:

```xml
<?xml version="1.0" encoding="utf-8"?>
<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema">
  <xs:element name="Weavers">
    <xs:complexType>
      <xs:all>
        <!-- Insert xs:element here -->
      </xs:all>
      <xs:attribute name="VerifyAssembly" type="bool" />
      <xs:attribute name="VerifyIgnoreCodes" type="xs:string" />
    </xs:complexType>
  </xs:element>
  <xs:simpleType name="bool">
    <xs:restriction base="xs:string">
      <xs:enumeration value="true" />
      <xs:enumeration value="false" />
    </xs:restriction>
  </xs:simpleType>
  <!-- Insert additional definitions here -->
</xs:schema>
```

(documentation omitted for brevity)

A requirement to prefix names with the weaver name (like in the example) may be a good idea. The file extension could be changed from xcf to xsd.

I thought about using `xs:include`, but it would either require to copy all xsd files to the project directory or to point to files in the nuget cache, and none of these solutions seemed good to me.

What do you think?

Again, sorry for the late suggestions @tom-englert. If you like the idea but don't want to change your implementation _yet again_ I will happily make these changes, just tell me.

ltrzesniewski:
OK then :+1:

I thought about the type conflicts and hence the requirement to prefix all type names with the weaver name, but you're right about the complexity this would create. Let's keep it simple.

tom-englert:
Anyway I don't expect many additional weavers adding support for this. 
I already added it to the most prominent ones (PropertyChanged, Costura, NullGuard)
@SimonCropp: is there any other of your weavers that uses configuration and you would say it's worth adding this?

SimonCropp:
I like the idea of getting intelisense when editing the configuration. Also it will prevent bugs where people have a typo in an attribute name

So happy for me to merge?

ltrzesniewski:
LGTM :+1:

SimonCropp:
Ok this is now deployed

SimonCropp:
Thanks for the significant work that went into this

ltrzesniewski:
Looks like `FodyWeavers.xsd` made its way to the file generated by `dotnet new gitignore`, FYI.

PR #602, opened 2018-11-20
Fix #601: remove VS2017 specific item

### LINKED ISSUE #601: Visual Studio 2015: Unknown attribute 'Update' in <None> element of Fody.targets
#### Describe the bug

Can not load project in Visual Studio 2015 after updating to latest version 3.3.0.
When running VS with `set TRACEDESIGNTIME=true`, the following error is shown (in German):
> xyz.csproj : error  : Das Update-Attribut im <None>-Element ist unbekannt.  xyz\packages\Fody.3.3.0\build\Fody.targets

which roughly translates to

> xyz.csproj : error  : The Update attribute in the <None> element is unknown.  xyz\packages\Fody.3.3.0\build\Fody.targets

#### Expected behavior

I feel that the latest library version should at least support the previous Visual Studio version 2015 as not everyone can afford to always use the latest version of Visual Studio.

tom-englert:
Fix #601: remove VS2017 specific item
This was just a "nice to have" feature, users can easily add this by their own.

### DIFF
--- a/Fody.sln
+++ b/Fody.sln
@@ -25,6 +25,7 @@ EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution Items", "{CDCDDCD6-6958-4352-AF0C-C589ECD51C0D}"
 	ProjectSection(SolutionItems) = preProject
 		Directory.Build.props = Directory.Build.props
+		README.md = README.md
 	EndProjectSection
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "RepoSync", "RepoSync\RepoSync.csproj", "{8C9C3B12-7F35-4E63-A230-E05F2F306A3D}"
--- a/Fody/Fody.targets
+++ b/Fody/Fody.targets
@@ -15,12 +15,6 @@
     <None Include="@(WeaverFiles)" />
   </ItemGroup>
 
-  <ItemGroup>
-    <None Update="$(MSBuildProjectDirectory)\FodyWeavers.xsd">
-      <Visible>false</Visible>
-    </None>
-  </ItemGroup>
-
   <UsingTask TaskName="Fody.WeavingTask" AssemblyFile="$(FodyAssembly)" />
   <UsingTask TaskName="Fody.UpdateReferenceCopyLocalTask" AssemblyFile="$(FodyAssembly)" />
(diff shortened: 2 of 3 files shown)


ltrzesniewski:
Are you sure we want to support old stuff forever? VS2019 is around the corner...
People using old VS versions can just keep using an old Fody version as well IMO.

SimonCropp:
i am on the fence on this one. yes, in this case, it is low cost. But it sets up an expectation. Also we have no idea what changes in the future will break VS 2015. we dont run unit tests or integration tests on VS2015. We dont smoke test on VS2015. If someone raises a non-trivial bug against VS2015 i dont want to required to install VS2015 or spin up a new VM with it.

I also think it is unreasonable for someone consuming an OSS product to expect to be able to anchor one set of technologies in the past, and continually expect to be able to use the newest of all other technologies. They have incurred technical debt, and seem to think it is OK for OSS projects to help keep paying the interest on that debt

If it was a paid product it would be a completely different matter.

thoughts?

ltrzesniewski:
FWIW, for this specific item I think we can keep VS2015 compatibility while retaining the VS2017 feature by adding the following property:

```xml
<DefaultItemExcludes>$(DefaultItemExcludes);FodyWeavers.xsd</DefaultItemExcludes>
```

I haven't tested it, but it should work. Which should make this _"zero cost"_ instead of low cost.

I still think that if you use an old IDE you shouldn't expect the newest package versions to work. Just stick with a previous version which works for you. Especially if no one tests new releases on VS2015.

It's your call @SimonCropp.. Either way maybe a "Supported IDE" section in the readme would be a good addition.

tom-englert:
About supported IDE: it's already there, just need to be merged: https://github.com/tom-englert/Fody/wiki/SupportedRuntimesAndIde

IMO we should try keeping backward compatibility as long as possible. It's always very frustrating when you have build all your code around a tool, and that suddenly stops working.
Of course you can rollback and use an older version, but what if you need a fix or a new feature?

AppVeyor has all the old Visual Studios available, so I'll try to setup a small test project that shows builds with older versions.

ltrzesniewski:
@tom-englert could you add my `DefaultItemExcludes` to the PR please? I can't provide a suggestion through GitHub for a line which is outside of the edited code area, and I'd prefer the next release to behave just like the last one.

About the supported IDE: Shouldn't this be more prominent? Supported runtimes are already in the readme.

ltrzesniewski:
Thanks! 😄

Issue #605, opened 2018-11-21
Remove Symbol loading for weaver addins


SimonCropp:
Weavers should move to embedded symbols to get line numbers for exceptions

PR #606, opened 2018-11-24
Cleanup FodyPackaging.targets

ltrzesniewski:
This just does some cleanup in FodyPackaging. I figured out some stuff could be done there while troubleshooting a packaging issue I had.

Here are the changes:
 - Replaced the obsolete `CreateItem` task usage with an `ItemGroup`
 - Removed pdb files from the package because of #605 
 - Removed an unnecessary file copy
 - Renamed `Weaver.Props.Template` to `Weaver.props` to achieve the previous point (the source file extension needs to match the NuGet target path extension in order to avoid the copy).  
IMO it's beneficial: it's a props file after all, and you get proper syntax highlighting in your editor when you open it.
 - The error checks now *only* check for the dll. Previously they could succeed if the xcf file was there but not the dll (which would be improbable but still possible).


ltrzesniewski:
I also added a commit that sets `DebugType` to `embedded` by default for weaver projects, due to #605.

tom-englert:
:+1:  works great!

SimonCropp:
awesome

SimonCropp:
this is now deployed

PR #615, opened 2018-11-28
Add an option to opt out of xsd generation

### LINKED ISSUE #613: FodyWeavers.xml warnings
Hello. I just upgraded from Fody version 3.2.4 to 3.3.2 and now I'm getting two new warnings:
> The element 'Weavers' cannot contain child element 'Costura' because the parent element's content model is empty.

> The element cannot contain white space. Content model is empty.

The first one happens for every weaver in any project. The second one only happens for projects where Fody is installed but no weaver is configured. I have a solution with multiple projects, but not all of them actually have weavers configured. I chose to install Fody in all of them for consistence reasons.

I think this is happening because of the new .xsd file, but I have zero knowledge about that stuff. I had to change the .xml files for two reasons:
1) To keep only one .xsd file for the whole solution. I temporarily rolled that back in order to make sure that this wasn't the issue cause.
2) The package didn't automatically changed my .xml files which are not named `FodyWeavers.xml`, which I use to deploy different weavers in debug and release mode, base on [this solution](https://github.com/Fody/Costura/issues/143#issuecomment-307963083).

ltrzesniewski:
Closes #613.

### DIFF
--- a/Fody/ConfigFileFinder.cs
+++ b/Fody/ConfigFileFinder.cs
@@ -7,8 +7,8 @@ using Fody;
 
 public static class ConfigFile
 {
-    static XNamespace schemaNamespace = XNamespace.Get("http://www.w3.org/2001/XMLSchema");
-    static XNamespace schemaInstanceNamespace = XNamespace.Get("http://www.w3.org/2001/XMLSchema-instance");
+    static readonly XNamespace schemaNamespace = XNamespace.Get("http://www.w3.org/2001/XMLSchema");
+    static readonly XNamespace schemaInstanceNamespace = XNamespace.Get("http://www.w3.org/2001/XMLSchema-instance");
 
     public static List<string> FindWeaverConfigs(string solutionDirectoryPath, string projectDirectory, ILogger logger, IEnumerable<string> wellKnownWeaverFiles)
     {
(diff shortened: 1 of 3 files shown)


ltrzesniewski:
Hmmm... Good idea, that could be a nice feature.

But what should happen if someone says `GenerateXsd="true"` in the solution-level file? Should we just ignore it, or throw an error? Currently no xsd can be generated solution-wide because we don't have the list of weavers.

BTW I discovered Fody supports solution-wide configs while doing this PR, I don't see it being documented anywhere.

tom-englert:
`GenerateXsd` should be true by default also on solution level. 
However this does not mean a schema file will be generated for the global `FodyWeavers.xml`, but only controls the generation for the individual `FodyWeavers.xml` files.

tom-englert:
On the other hand, if this is no longer documented anywhere, maybe @SimonCropp wanted this to be a deprecated feature?

tom-englert:
In combination with #617 this should be really sufficient as it is :+1:

ltrzesniewski:
@tom-englert ok, but I wouldn't be surprised if this PR conflicts with #617. How do you want to merge these?

tom-englert:
If everyone agrees, I can merge yours first, and then re-base mine.

SimonCropp:
no objection here

SimonCropp:
Re solution level weavers. Do u think i would make things easier if we removed it? If so i am happy to schedule if for the next major? But if so we should add a warning when we detect one now

ltrzesniewski:
If you think they are valuable, then IMO they certainly don't need to be removed because of IntelliSense support (which is essentially a convenience feature). I can't really tell how much they are valuable to users.

Thinking about it, I'd rather add support for a `<FodyGenerateXsd>false</FodyGenerateXsd>` MSBuild property which could be added to a `Directory.Build.props` to turn off xsd support entirely for a scope.

I'll either add it to this PR or make another PR depending on when @tom-englert merges this one.

PR #617, opened 2018-12-01
#613: Defer schema update so legacy weavers are included, too.

tom-englert:
this should improve experience with legacy weavers (#613), they are now included in the schema, too.


Fody/ConfigFileFinder.cs
```diff
@@ -73,21 +73,32 @@ static void GenerateDefault(string projectConfigFilePath, IEnumerable<string> we
...
-    static void CreateSchemaForConfig(string projectConfigFilePath, IEnumerable<string> wellKnownWeaverFiles)
+    static string GetWeaverName(WeaverEntry weaver)
     {
-        if (wellKnownWeaverFiles == null)
+        return string.IsNullOrEmpty(weaver.Element) ? weaver.TypeName : weaver.AssemblyName;
+    }
+
+    static void CreateSchemaForConfig(string projectConfigFilePath, IEnumerable<string> wellKnownWeaverFiles, IEnumerable<WeaverEntry> weavers)
+    {
+        var weaverFiles = weavers?.ToDictionary(GetWeaverName, weaver => weaver.AssemblyPath, StringComparer.OrdinalIgnoreCase)
```
ltrzesniewski: I can't really review this properly right now because I only have my tablet, but I'd avoid `ToDictionary` here because we *could* get a duplicate key. 
Maybe wrapping the whole method in a `try` block would be a good idea. If anything goes wrong here, emit a warning instead of a build error.

ltrzesniewski:
Sorry, I haven't noticed there's _already_ a `try` block there. It's easier to see on an actual computer 😉 

Anyway I tested it, and it looks good :+1:

Legacy weavers are included in the xsd only after the project is built with the weaver already in the xml file, but I think that behavior is OK for legacy scenarios. I also tested it on an in-solution weaver and everything works fine.

@tom-englert I'll let you merge both branches as you suggested in #615.

SimonCropp:
Let me know when u need me to do a release

PR #618, opened 2018-12-01
Fix #616

### LINKED ISSUE #616: Debugging weavers no longer possible
since default debug type has changed to "embedded" in https://github.com/Fody/Fody/commit/e924e1f361ab36fd37b5dff77561965285f05eae I can't debug my weavers any longer. 
In the modules view it says "xyz.Fody - Binary was not built with debug information".
Setting debug type explicitly to e.g. "portable" fixes the problem.

Any thoughts? Are there additional settings needed so that VS can load embedded symbols?

tom-englert:
Fix #616 
Revert https://github.com/Fody/Fody/commit/2a75c613d44dd278474cf335dec3780c25dc1b8d
Embedded symbols don't work this way (MS bug)

### DIFF
--- a/FodyIsolated/AssemblyLoader.cs
+++ b/FodyIsolated/AssemblyLoader.cs
@@ -21,9 +21,22 @@ public partial class InnerWeaver
         return assemblies[assemblyPath] = loadFromFile;
     }
 
-    public static Assembly LoadFromFile(string assemblyPath)
+    static Assembly LoadFromFile(string assemblyPath)
     {
         var rawAssembly = File.ReadAllBytes(assemblyPath);
+
+        var pdbPath = Path.ChangeExtension(assemblyPath, "pdb");
+        if (File.Exists(pdbPath))
+        {
+            return Assembly.Load(rawAssembly, File.ReadAllBytes(pdbPath));
+        }
+        
+        var mdbPath = Path.ChangeExtension(assemblyPath, "mdb");
+        if (File.Exists(mdbPath))
+        {
+            return Assembly.Load(rawAssembly, File.ReadAllBytes(mdbPath));
+        }
+
         return Assembly.Load(rawAssembly);
     }
 }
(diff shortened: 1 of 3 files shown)


FodyIsolated/AssemblyLoader.cs
```diff
@@ -21,9 +21,22 @@ public Assembly LoadWeaverAssembly(string assemblyPath)
...
     {
         var rawAssembly = File.ReadAllBytes(assemblyPath);
+
+        var pdbPath = Path.ChangeExtension(assemblyPath, "pdb");
+        if (File.Exists(pdbPath))
+        {
+            return Assembly.Load(rawAssembly, File.ReadAllBytes(pdbPath));
+        }
+        
+        var mdbPath = Path.ChangeExtension(assemblyPath, "mdb");
```
ltrzesniewski: I think you can safely drop the mdb part. It's not included by FodyPackaging, and who builds Fody addins with mono anyway?

tom-englert: Every time we remove something that we think nobody is using, we end up with new issues :smirk: 
The packaging part can be easily customized, the weaver not...

SimonCropp:
do we have a reference to the MS bug?

tom-englert:
@SimonCropp not yet, still wondering where to post this issue - to me it looks rather like a framework bug than a VS bug. 
Where would you post this?

SimonCropp:
@terrajobst @davkean  thoughts on where to raise a bug with VS not picking up embedded symbols when loaded via a byte array?

tom-englert:
I have tried here:
https://developercommunity.visualstudio.com/content/problem/397080/assemblyloadloadbyte-does-not-load-embedded-smbols.html

tom-englert:
It's now https://developercommunity.visualstudio.com/content/problem/397436/assemblyloadloadbyte-does-not-load-embedded-smbols-2.html

PR #623, opened 2018-12-03
update open collective sponsor link

piamancini:
I think these two links are better link to send the companies to. We'll be shipping a new contribution flow soon and that wiki section will be deprecated.


SimonCropp:
@piamancini thanks for this. and your help in other open collective areas :)

Issue #625, opened 2018-12-04
Add configuration to turn of legacy weaver lookup


tom-englert:
Scanario: 
A weaver is configured in FodyWeavers.xml, but not included as NuGet package. 
The build may succeed, because Fody finds any weaver by legacy lookup. 

To make this more strict, there should be a flag in configuration that allows to turn of legacy lookup.
Alternative: Turn of legacy lookup by default, and have a flag to turn it on again.

@SimonCropp @ltrzesniewski Whats your opinion about this?

ltrzesniewski:
Seems reasonable. I wouldn't mind having legacy support off by default (most weavers should be up to date), as long as the error message is clear and references the config option.

SimonCropp:
Isnt legacy lookup prone to bugs via side effects? And if it doesn’t find one, isnt the fix trivial for the user?... add the package?

tom-englert:
Yeah, default option for legacy lookup should probably be OFF

SimonCropp:
What I don’t understand why would we want anyone to enable it?

tom-englert:
There might be some legacy weavers around, and it's not a big deal keeping it for a while...

SimonCropp:
Ok i am a little confused. What do u define as “legacy lookup” versus “legacy weaver”?

PR #631, opened 2018-12-16
Refactoring for #625

tom-englert:
This PR addresses #625 

After looking at the code I decided to not add a configurable option, but to refactor it an completely remove the legacy lookup.
The code flow was too much focused on the old strategy, where the `FodyWeavers.xml` was the "main entry point". Now that the `<WeaverFiles>` items are our starting point, many things could be simplified, so there is much less code left to maintain.

Currently there are some minor changes to the constraints, that make using weavers even easier and more flexible:

1. All weavers configured for a project (i.e. NuGet package installed) will run by default, even if there is no entry in the `FodyWeavers.xml`. It is no longer required (however still recommended), to list the weavers in the configuration.
2. A `FodyWeavers.xml` in the solution directory will control ordering and default configuration of the weavers. However now it's possible to configure the super set of all weavers used in all projects. Any additional configuration in the global configuration where no weaver is available in a specific project will be just ignored.
3. InSolutionWeaving with multiple weavers per project is still supported, however no longer by naming convention. The Wiki update that describes the steps needed for migration is prepared here: https://github.com/tom-englert/Fody/wiki/InSolutionWeaving


Fody/BuildLogger.cs
```diff
@@ -1,5 +1,7 @@
 using System;
+
```
SimonCropp: can u change whatever tool does this so it ignores the fody codebase?

SimonCropp:
I assume this is a breaking change?

Fody/WeaversHistory.cs
```diff
@@ -9,18 +9,22 @@ public static class WeaversHistory
     public static bool HasChanged(IEnumerable<string> weaverPaths)
     {
         var changed = false;
+
```
SimonCropp: while i am ok with some of the whitespace changes in this PR (eg this one). it would be less noisy to review this PR if the whitespace changes were done either before or after this PR

Fody/WeavingTask.cs
```diff
@@ -101,14 +102,21 @@ public override bool Execute()
             return success;
         }
 
-        List<string> GetWeaverFilesFromProps()
+        IEnumerable<WeaverEntry> GetWeaversFromProps()
         {
             if (WeaverFiles == null)
             {
-                return new List<string>();
+                return new WeaverEntry[0];
```
SimonCropp: minor but `Array.Empty`

Integration/SampleTargetTests/FodyWeavers.xsd
```diff
@@ -39,6 +38,11 @@
           <xs:documentation>A comma-separated list of error codes that can be safely ignored in assembly verification.</xs:documentation>
         </xs:annotation>
       </xs:attribute>
+      <xs:attribute name="GenerateXsd" type="xs:boolean">
```
SimonCropp: Should this be in this PR?

Integration/WithNoPdb/WithNoPdb.csproj
```diff
@@ -4,6 +4,6 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Fody" Version="$(IntegrationVersion)" />
-    <PackageReference Include="Virtuosity.Fody" Version="1.21.1" />
+    <PackageReference Include="Virtuosity.Fody" Version="1.21.3" />
```
SimonCropp: perhaps in a diff PR?

readme.md
```diff
@@ -307,7 +308,7 @@ Older versions of Visual Studio may still work, but are not actively supported.
 
 ## Supported Runtimes
 
- * Classic .NET: See *Support ended* in [NET Framework version history](https://en.wikipedia.org/wiki/.NET_Framework_version_history#Overview). i.e only 4.5.2 and above is supported.
+ * Classic .NET: See *Support ended* in [NET Framework version history](https://en.wikipedia.org/wiki/.NET_Framework_version_history#Overview). i.e only 4.5.2 and above is actively supported.
```
SimonCropp: diff commit?

readme.md
```diff
@@ -273,7 +273,8 @@ Install-Package Virtuosity.Fody
 
 ### Add FodyWeavers.xml
 
-To indicate what weavers run and in what order a `FodyWeavers.xml` file is used at the project level. It needs to be added manually. It takes the following form:
+To indicate what weavers run and in what order a `FodyWeavers.xml` file is used at the project level.
```
SimonCropp: diff commit?

Fody/BuildLogger.cs
tom-englert: Sorry, this was R#, I will include this setting in the .setting file.

Fody/WeaversHistory.cs
tom-englert: I will remove those, slipped in because BeyondCompare can hide all whitespace-only changes.

Fody/WeavingTask.cs
tom-englert: Fixed

Integration/SampleTargetTests/FodyWeavers.xsd
tom-englert: That's what the latest Fody generates...

Integration/WithNoPdb/WithNoPdb.csproj
tom-englert: I was not working properly with the old version, so I had to update here.

readme.md
tom-englert: agree

readme.md
tom-englert: agree

tom-englert:
@SimonCropp I've cleaned up the unnecessary whitespace changes introduced by R#, and extended the R# settings to avoid this in the future.
Also removed the readme changes, will post that later.

The breaking changes are the three listed in the top comment.
1. and 2. are up to discussion if we want to allow this extensions, or rather keep it like it was before.
3. Is mandatory, if we don't want to keep 90% of the old code.

ltrzesniewski:
I'll try to review this properly this evening, but here are some thoughts:

> 1. All weavers configured for a project (i.e. NuGet package installed) will run by default, even if there is no entry in the `FodyWeavers.xml`. It is no longer required (however still recommended), to list the weavers in the configuration.

IMO this is a huge change in behavior and I think the new behavior is confusing. What would you think about the following?
- If there is no `FodyWeavers.xml`, run all weavers. Make this file optional and don't create it if it doesn't exist.
- If there is a `FodyWeavers.xml`, keep the existing behavior (run only the weavers in the file)

I think most people use a single weaver with default config (I may be wrong, but I'm thinking about `PropertyChanged` here), so this would simplify Fody usage for a lot of people.

> 2. A `FodyWeavers.xml` in the solution directory will control ordering and default configuration of the weavers. However now it's possible to configure the super set of all weavers used in all projects. Any additional configuration in the global configuration where no weaver is available in a specific project will be just ignored.

I like that :+1:, but this absolutely *must* work if a project is compiled separately (without a solution) or we may end up with inconsistent results. In this case, should we look in the parent directory of the project, or do a lookup in all ancestor directories (like `Directory.Build.props`)?

> 3. InSolutionWeaving with multiple weavers per project is still supported, however no longer by naming convention. The Wiki update that describes the steps needed for migration is prepared here: https://github.com/tom-englert/Fody/wiki/InSolutionWeaving

I'd consider this as an advanced usage scenario, so I think this change is OK.

tom-englert:
_1: What about this: 
- If there is no `FodyWeavers.xml`, create it as it's done already now. 
- If you have a weaver installed that is not in `FodyWeavers.xml`, generate a warning...

_2: The configuration at solution level is already an  existing feature of Fody, and I think it will not work if you don't have a solution, i.e. compile just a single project. 

Considering this we may either drop it at all, or completely redesign it.

ltrzesniewski:
1. That would work too and would not be confusing thanks to the warning. But I thought that not requiring `FodyWeavers.xml` for simple cases would be a good idea. Which behavior do you prefer @SimonCropp ?

2. Currently, Fody uses the parent of the project directory if the solution path is not provided ([see here](https://github.com/Fody/Fody/blob/ca927ee7fef2918b8a8ae5eb36fa36a546e7ec2e/Fody/SolutionDirectoryFinder.cs#L17)), and I think that's better than not doing anything and having different compilation results depending on how the project is built.

> Considering this we may either drop it at all, or completely redesign it.

Are you talking about the solution-level file? Since it's not documented I'd say we can pretty much do anything we want with it. But I don't know if it was documented before.

Tests/Fody/ConfigFileFinderTests.cs
```diff
@@ -100,13 +102,13 @@ public void ShouldOptOutOfXsd()
 </Weavers>
 ");
 
-        var wellKnownWeaverFiles = new[] { @"something\TestWeaver.Fody.dll" };
+        var wellKnownWeaverFiles = new[] { new WeaverEntry { AssemblyPath = @"something\TestWeaver.Fody.dll" } };
```
ltrzesniewski: Rename `wellKnownWeaverFiles` to `weavers` for consistency like above.
Same comment for the two functions below.

Fody/FileChangedChecker.cs
```diff
@@ -1,12 +1,12 @@
 public partial class Processor
 {
-    public virtual bool ShouldStartSinceFileChanged()
+    public virtual bool TargeAssemblyHasAlreadyBeenProcessed()
```
ltrzesniewski: Typo

Tests/Fody/ShouldStartSinceFileChangedTests.cs
```diff
@@ -7,26 +7,28 @@ public class ShouldStartSinceFileChangedTests : TestBase
...
-            {
-                Logger = new Mock<BuildLogger>().Object,
-                AssemblyFilePath = typeof (string).Assembly.CodeBase.Replace("file:///", "")
-            };
-        Assert.True(processor.ShouldStartSinceFileChanged());
+        {
+            Logger = new Mock<BuildLogger>().Object,
+            AssemblyFilePath = typeof(string).Assembly.CodeBase.Replace("file:///", "")
+        };
+        Assert.False(processor.TargeAssemblyHasAlreadyBeenProcessed());
```
ltrzesniewski: This file needs to be renamed

Fody/ConfigFileFinder.cs
```diff
@@ -3,98 +3,116 @@
...
+
+public class WeaverConfigEntry
+{
+    public WeaverConfigFile ConfigFile;
+    public string ElementName;
+    public string Content;
+    public int ExecutionOrder;
+}
 
 public static class ConfigFile
```
ltrzesniewski: This file needs to be renamed to `ConfigFile.cs`. Not sure if `WeaverConfigFile` and `WeaverConfigEntry` should stay here or if they deserve separate files.

Fody/Processor.cs
```diff
@@ -59,71 +58,73 @@ public virtual bool Execute()
...
         {
+            stopwatch.Stop();
             Logger.LogInfo($"  Finished Fody {stopwatch.ElapsedMilliseconds}ms.");
         }
     }
 
     void Inner()
     {
-        ValidateProjectPath();
+        if (Weavers.Count == 0)
```
ltrzesniewski: Old style in-solution weavers now give me a "no weavers found" warning instead of complaining that my `FodyWeavers.xml` references an unknown weaver.
I'd say this check is made too early.

Integration/SampleTargetTests/FodyWeavers.xsd
SimonCropp: yeah i was just trying to minimise the noise in this pr

Tests/Fody/ShouldStartSinceFileChangedTests.cs
tom-englert: Yes, there are a few files that need this, however I would like to do this in a separate PR, since changing content + renaming often kills the history...

Fody/ConfigFileFinder.cs
tom-englert: S.A.

FodyIsolated/InnerWeaver.cs
```diff
@@ -184,7 +184,7 @@ void ExecuteWeavers()
                 {
                     throw new Exception($"Failed to execute weaver {weaver.Config.AssemblyPath}", exception);
                 }
-                var finishedMessage = $"  Finished '{weaver.Config.AssemblyName}' in {startNew.ElapsedMilliseconds}ms {Environment.NewLine}";
+                var finishedMessage = $"  Finished '{weaver.Config.AssemblyBaseName}' in {startNew.ElapsedMilliseconds}ms {Environment.NewLine}";
```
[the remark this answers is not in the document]
tom-englert: Right, I've missed those.

Fody/WeaversXmlHistory.cs
```diff
@@ -1,38 +1,29 @@
-using System;
 using System.Collections.Generic;
-using System.IO;
+using System.Linq;
 
-public partial class Processor
+public static class WeaversConfigHistory
 {
-    public static Dictionary<string, DateTime> TimeStamps = new Dictionary<string, DateTime>();
+    static string WeaverConfigsContent;
```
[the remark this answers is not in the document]
tom-englert: As long as it runs again in the same MSBuild process, it works; however I think most of the time it just slips this. 
I wonder if it's worth fixing this, or if we just should drop it at all.

Fody/Processor.cs
```diff
@@ -59,71 +58,73 @@ public virtual bool Execute()
...
-        ConfigureWhenWeaversFound();
+                return;
+            }
+        }
 
-        ConfigureWhenNoWeaversFound();
+        ConfigFile.EnsureSchemaIsUpToDate(ProjectDirectory, Weavers, GenerateXsd);
 
-        ConfigFile.EnsureSchemaIsUpToDate(ProjectDirectory, WeaverFilesFromProps, Weavers, GenerateXsd);
+        lock (mutex)
```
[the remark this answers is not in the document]
tom-englert: Maybe MSBuild could start multiple threads? That's the only scenario I could think of.

Fody/ConfigFileFinder.cs
```diff
@@ -3,98 +3,116 @@
...
     static readonly XNamespace schemaInstanceNamespace = XNamespace.Get("http://www.w3.org/2001/XMLSchema-instance");
 
-    public static List<string> FindWeaverConfigs(string solutionDirectoryPath, string projectDirectory, ILogger logger, IEnumerable<string> wellKnownWeaverFiles, bool defaultGenerateXsd)
+    public static IEnumerable<WeaverConfigFile> FindWeaverConfigFiles(string solutionDirectoryPath, string projectDirectory, ILogger logger)
     {
-        var files = new List<string>();
-
-        var solutionConfigFilePath = Path.Combine(solutionDirectoryPath, FodyWeaversConfigFileName);
-        if (File.Exists(solutionConfigFilePath))
         {
```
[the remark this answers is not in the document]
tom-englert: yes, to prohibit use of `solutionConfigFilePath` in the lower block. I'm not fully happy with this code, but for now it just works..

tom-englert:
@ltrzesniewski you could easily ensure that `SolutionDir` is properly set be checking it in your `Directory.Build.props`, if this is at solution level.

ltrzesniewski:
What if you don't *want* a solution at all?
In some cases I run MSBuild directly on a project file in CI builds, and I don't want to have to configure the solution path for each solution just to make Fody happy.
I think the current "assume it's the parent directory" approach is fine.

tom-englert:
@ltrzesniewski whatever works for you :smile:

tom-englert:
UTC+1

ltrzesniewski:
@tom-englert yeah I don't use solution-level config at all, I just want to make sure Fody keeps working fine if there's no solution. Your PR didn't break that when I tested it 😉 

@SimonCropp Also UTC+1 here, although I'm not sure I'll have much time before I go on vacation. Although slower, text communication is "async" so it would work better for me ATM. Also I'm not sure I have much more to add here.

tom-englert:
From my POV this is now final. 
The behavior changes are as follows:
1. Not changed: Only weavers listed in `FodyWeavers.xml` will be run.
2. New: Weavers installed via NuGet but not listed in  `FodyWeavers.xml` generate a warning.
3. Changed: If there is no  `FodyWeavers.xml`, one is generated, but now only with a warning instead of an error.
4. Changed: Extra entries in a solution-level `FodyWeavers.xml` are ignored, so you can now configure the superset of all weavers in the solution.
5. InSolutionWeavers are no longer detected by convention but must be declared explicitly.
6. Legacy weavers that do not provide a .props file with a `<WeaverFiles>` entry are no longer supported.

Only 5. is a breaking change. You will get an error message that points you to the wiki where it will be documented.

A solution level  `FodyWeavers.xml` has never been documented explicitly, so 4. is IMO not breaking.

Any objections to merge this?

SimonCropp:
Sorry i havnt had a chance to properly review this. I will get to it tomorrow

SimonCropp:
Given solution level config was never documented, and this is already a breaking change, thoughts on properly removing support for solution level config?

tom-englert:
@SimonCropp with the new flexibility, a solution level config will be a good feature, and the additional code needed is very limited. I will add some documentation if we release it like this.

It is not really a breaking change. If s.o. was using this, a successful build will build the same as before, only a build that failed before might succeed now - so it's rather an unbreaking change :smile:

SimonCropp:
@tom-englert 

re "It is not really a breaking change."

How is the below not a breaking change?

> Legacy weavers that do not provide a .props file with a <WeaverFiles> entry are no longer supported.

tom-englert:
@SimonCropp "not really a breaking change" was of course only related to the behavior adjustment of the solution level configuration.

SimonCropp:
Have u had a look at how many weavers need to be moved over to props?

tom-englert:
Just made some statistics, based on the weavers listed in the https://github.com/Fody/Fody/blob/master/readme.md:

https://1drv.ms/x/s!Auam3jxFhE-WgeAmQCSEAK9Ynl28PQ

74 Weavers listed
40 Up-to-date
25 Not even updated to Fody 3.x

9 Updated to Fody 3.x, but do not provide props:

Weaver | Url
-- | --
WeakEvents | https://github.com/adbancroft/WeakEvents.Fody
Catel | https://github.com/Catel/Catel.Fody
Tracer | https://github.com/csnemes/tracer
SwallowExceptions | https://github.com/duaneedwards/SwallowExceptions
Undisposed | https://github.com/ermshiperete/undisposed-fody
Vandelay | https://github.com/jasonwoods-7/Vandelay
Realm | https://github.com/realm/realm-dotnet/tree/master/Weaver/RealmWeaver.Fody
MethodBoundaryAspect | https://github.com/vescon/MethodBoundaryAspect.Fody
Bindables | https://github.com/yusuf-gunaydin/Bindables

tom-englert:
I have created an issue in these 9 projects.

SimonCropp:
@tom-englert i applied the update here https://github.com/tom-englert/Fody/wiki/InSolutionWeaving  but i called it v4. i created a milestone https://github.com/Fody/Fody/milestone/101  i will start fleshing out the descriptions of the issues. i also pushed a beta nuget of v4 if u want to play with it

SimonCropp:
@tom-englert can u invoice $100 for this

tom-englert:
@SimonCropp :smile: Done

BrunoJuchli:
@tom-englert 
I expected this change to result in a single FodyWeavers.xml on solution level to suffice (no more FodyWeavers.xml per project). 

However I get
>MSBUILD : error : Fody: Failed to update schema for (C:\work\FodySolutionConfigRepro\Lib\FodyWeavers.xml). Exception message: Could not find file 'C:\work\FodySolutionConfigRepro\Lib\FodyWeavers.xml'.

when building in VS.
(I also tried setting `FodyGenerateXsd` to false @ lib.csproj and `GenerateXsd` to false @ FodyWeavers.xml in the root.)

Before I dig into analyzing the cause and providing a fix: is this even what you had in mind with the solution level config? (The wiki doesn't mention this config option).

PR #636, opened 2018-12-29
Build failure when no FodyWeavers.xml at project level

nexussays:
I was updating various build components to their latest versions and when updating Fody encountered a breaking change I believe was made here https://github.com/Fody/Fody/pull/598/files#diff-46dbf35f7a743d6892884d244ffd5d5fR28 which changed a `logger.LogDebug` to `logger.LogError` and seems to be causing builds to fail.

This PR has a fix implementing what I believe is the desired logic: if a config file is found in the solution directory, not finding a config file in the project directory won't log an error and fail a running build.

I did read the guidelines in the PR template but hope this small fix is acceptable. Cheers.

### DIFF
--- a/Fody/ConfigFileFinder.cs
+++ b/Fody/ConfigFileFinder.cs
@@ -25,19 +25,22 @@ public static class ConfigFile
         var projectConfigFilePath = Path.Combine(projectDirectory, FodyWeaversConfigFileName);
         if (!File.Exists(projectConfigFilePath))
         {
-            if (!files.Any() && wellKnownWeaverFiles != null)
+            if (!files.Any())
             {
-                GenerateDefault(projectConfigFilePath, wellKnownWeaverFiles, defaultGenerateXsd);
+                if (wellKnownWeaverFiles != null)
+                {
+                    GenerateDefault(projectConfigFilePath, wellKnownWeaverFiles, defaultGenerateXsd);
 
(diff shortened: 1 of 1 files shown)


tom-englert:
@nexussays since you are using a solution level config file, what do you think about the enhancements to the solution level configuration discussed in #631

ltrzesniewski:
But won't this change become instantly obsolete because of #631?

tom-englert:
@ltrzesniewski yes, but will #631 be merged in time?

nexussays:
@tom-englert  In my reading of #631 it doesn't seem like my use-case will be affected: I have a single `FodyWeavers.xml` at the solution-level and an accompanying `Directory.Build.props` to add the corresponding packages. I don't customize anything per-project. Thanks!

SimonCropp:
@nexussays sorry for the hassle. but can you rebase this

tom-englert:
@nexussays @SimonCropp I think #631 has made this already obsolete

nexussays:
@SimonCropp Tried to rebase; the code has been refactored and it seems this fix is no longer needed. Are those changes going to be deployed soon? Cheers!

SimonCropp:
There is a beta deployed to nuget.

SimonCropp:
There is some minor breaking changes. So some doco/guidance is required

SimonCropp:
ok v4 is release. so this should be fixed

PR #650, opened 2019-01-12
Register the references cache file

ltrzesniewski:
Minor cleanup: This causes the cache file to be deleted when the Clean target is called.


ltrzesniewski:
I changed the implementation to add the item in `FodyTarget`, where the file is actually written. MSBuild output inference is used when `FodyTarget` is up to date.

PR #651, opened 2019-01-12
Add up-to-date check for FodyWeavers.xml

ltrzesniewski:
This adds FodyWeavers.xml to the list of files that will be checked to determine if the project is up-to-date in the new project system.

This is documented [here](https://github.com/dotnet/project-system/blob/master/docs/repo/up-to-date-check.md).


SimonCropp:
awesome. thanks

PR #654, opened 2019-01-13
Update readme.md

AmazingTurtle:
Use correct link to addins list

Where it is clear the below content has not read, the issue is likely to be closed with "please read the template". Please don't take offense at this. It is simply a time management decision. When someone raises an issue, without reading the template, then often too much time is spent going back and forth to obtain information that is outlined below.

#### Description

Link to addin list was dead.

#### The solution

Updated the link to the correct one.

#### Todos

-none-


SimonCropp:
Thanks

PR #660, opened 2019-01-28
Simplify FodyPackaging

SimonCropp:
This is a breaking change to FodyPackaging.

Some history: FodyPackaging was originally created to only build packages from within the Fody Org. As such many things were hard coded (or specified with a rigid convention) eg license, icon, output dir, project url. FodyPackaging was then made more flexible so it could target projects outside the Fody Org. This was a good change, but had some side effects

 * to achieve this significant complexity had to be added to make the above conventions  overridable or tweaked. 
 * pulled in the assumption/convention that all fody addins are hosted on github.
 * complexity and fragility around determining the output path

Moving forward i believe this will be difficult to document and to maintain

This PR remove those conventions that are complex or overly presumptuous.

All downstream weavers that consume FodyPackaging would need to set their own

 * PackageOutputPath 
 * PackageLicenseExpression
 * PackageProjectUrl 
 * PackageIconFile

### DIFF
--- a/FodyPackaging/build/FodyPackaging.props
+++ b/FodyPackaging/build/FodyPackaging.props
@@ -2,10 +2,6 @@
   <PropertyGroup>
     <PackageId Condition="'$(PackageId)' == ''">$(MSBuildProjectName).Fody</PackageId>
     <PackageTags Condition="'$(PackageTags)' == ''">ILWeaving, Fody, Cecil, AOP</PackageTags>
-    <FodySolutionDir Condition="'$(FodySolutionDir)' == '' AND '$(SolutionDir)' != '*Undefined*'">$(SolutionDir)</FodySolutionDir>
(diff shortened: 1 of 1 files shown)


SimonCropp:
ping @glennawatson

SimonCropp:
ping @jasonwoods-7 @robinzevallos @wazowsk1 @GeertvanHorrik @vescon

GeertvanHorrik:
Fine by me, already using my own packaging mechanism. Why not create docs for the real minimum requirements like the code below:

https://github.com/Catel/Catel.Fody/blob/develop/src/Catel.Fody.Attributes/Catel.Fody.Attributes.csproj#L31

Then everyone can use the methods they prefer.

SimonCropp:
@GeertvanHorrik yep working on proper doco ATM

glennawatson:
For those using Azure DevOps/Pipelines, make sure you have latest .NET SDK 2.1.x, MSBuild picks up the SDK with the LicenseExpression from there.

PR #677, opened 2019-03-18
Allow ProjectWeaverXml to be set externally

redwyre:
#### Description

Allows ProjectWeaverXml to be set externally

#### The solution

Only set ProjectWeaverXml when it hasn't been set already
Allows doing something like:

```
<ProjectWeaverXml Condition="'$(Configuration)' == 'Release'">$(ProjectDir)FodyWeavers.xml</ProjectWeaverXml>
<ProjectWeaverXml Condition="'$(Configuration)' == 'Debug'">$(ProjectDir)FodyWeavers.xml</ProjectWeaverXml>
```

#### Todos

 * [ ] Related issues
 * [ ] Tests
 * [ ] Documentation

### DIFF
--- a/Fody/Fody.targets
+++ b/Fody/Fody.targets
@@ -1,7 +1,7 @@
 ﻿<Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
   <PropertyGroup>
-    <ProjectWeaverXml>$(ProjectDir)FodyWeavers.xml</ProjectWeaverXml>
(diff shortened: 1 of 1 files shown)


SimonCropp:
thanks for this 

This is now deployed in version 4.1.0-beta.7. NuGet may take some time to make it available for download.

redwyre:
@SimonCropp This doesn't work, the actual filename is hardcoded, I was going to look into what's required to pass it though

Issue #687, opened 2019-04-07
Simplify symbol loading


tom-englert:
I found that the "WithNoPdb" test project actually did generate a pdb file, and there was no test verifying this. 
This was probably due to changes in the MSBuild project structure controlling the generation of debug symbols.
Now Fody itself relies on some special MSBuild keywords to determine the kind of symbols generated, which might also change in future, so eliminating this dependency seems to be an improvement.

The PR will: 
- Add some test that verifies correct loading of symbols
- Fix "WithNoPdb" project to build without symbols
- Simplify symbol loading in Fody to remove the dependency on the MSBuild variable.

BrunoJuchli:
PR: #688

PR #688, opened 2019-04-07
#687

tom-englert:
- Add some test that verifies correct loading of symbols
- Fix "WithNoPdb" project to build without symbols
- Simplify symbol loading in Fody to remove the dependency on the MSBuild variable.


SampleWeaver.Fody/ModuleWeaver.cs
```diff
@@ -58,6 +61,44 @@ public override void Execute()
...
     }
 
+    void VerifySymbols()
+    {
+        LogInfo("Verify Symbols");
+
+        var anyMethod = ModuleDefinition.GetTypes()
+            .Where(type => type.IsClass)
+            .SelectMany(type => type.GetMethods())
+            .FirstOrDefault();
```
ltrzesniewski: You may be unlucky and end up with a compiler-generated method with no symbols. I'd check the following:
- The method is `!IsSpecialName && !IsRuntimeSpecialName` (to discard `.ctor`, `.cctor` etc)
- The type is not compiler-generated (maybe a check for non-word characters in its name could be sufficient)
- The method is not compiler-generated

SampleWeaver.Fody/ModuleWeaver.cs
```diff
@@ -58,6 +61,44 @@ public override void Execute()
...
+
+        if (anyMethod == null)
+        {
+            LogInfo("Assembly has no type with a method, symbol verifying skipped");
+            return;
+        }
+
+        var shouldHaveSymbols = ModuleDefinition.Assembly.CustomAttributes.All(attr => attr.AttributeType.Name != "NoSymbolsMarkerAttribute");
```
ltrzesniewski: Minor: A config attribute in the weaver would be a little cleaner IMO.

SampleWeaver.Fody/ModuleWeaver.cs
```diff
@@ -58,6 +61,44 @@ public override void Execute()
...
+        var hasSymbols = HasSymbols(anyMethod);
+        LogInfo("Assembly has symbols: " + hasSymbols);
+
+        if (shouldHaveSymbols != hasSymbols)
+        {
+            LogError($"Unexpected symbols in assembly {ModuleDefinition.FileName}, should have: {shouldHaveSymbols}, but has: {hasSymbols}");
+        }
+    }
+
+    bool HasSymbols(MethodDefinition method)
```
SimonCropp: can this be done without the try catch?

SimonCropp: bump

tom-englert: @SimonCropp what do you mean with bump? still any objections?

BrunoJuchli:
This relates to #687 

(I mention this because one has to write it in a text body for github to pick it up, and in #687 I read about a PR but had to go looking for it).

SimonCropp:
there were some edge cases where pdb/mdbs did not get written to the location cecil expected. i think windows phone was one. but i think those cases dont exist anymore. i will merge this and release a beta so i can verify on downstreams

FodyIsolated/ModuleReader.cs
```diff
@@ -1,46 +1,44 @@
...
-        SymbolStream?.Dispose();
-        File.Delete(tempAssembly);
-        if (File.Exists(tempSymbols))
+        var hasSymbols = false;
+        try
+        {
+            module.ReadSymbols();
+            hasSymbols = true;
+        }
+        catch
```
SimonCropp: is there a unit test that hits this exception?

tom-englert: Yes, added a test for no symbols:
https://github.com/Fody/Fody/pull/688/commits/dc8386f9c60b60586656c7b0acce25c22a07ea5c

FodyIsolated/ModuleReader.cs
```diff
@@ -1,46 +1,44 @@
...
-    public virtual void ReadModule()
+    public void ReadModule()
     {
-        tempAssembly = $"{AssemblyFilePath}.tmp";
-        File.Copy(AssemblyFilePath, tempAssembly, true);
-
-        if (debugReaderProvider != null && DebugSymbols != DebugSymbolsType.Embedded)
+        var result = ReadModule(AssemblyFilePath,assemblyResolver);
+        hasSymbols = result.hasSymbols;
+        if (hasSymbols)
```
tom-englert: @SimonCropp shouldn't that rather be if `(!hasSymbols)`?

FodyIsolated/ModuleReader.cs
tom-englert: Reading the history again I think I fixed the integration tests, and you added a unit test.

FodyIsolated/ModuleReader.cs
SimonCropp: `if (hasSymbols)` seems correct to me. since if it has symbols it used to duplicate the symbols file to prevent it being locked on read

tom-englert: Then I don't understand the message - what is the meaning of the log message "Debug symbols disabled." when the assembly has debug symbols?

SimonCropp: `Debug symbols disabled` was added in this PR. so i am not sure why

tom-englert: So the condition had really toggled during refactoring. Fixed it an improved the message text.

PR #724, opened 2019-07-20
Solution-wide config support for `EnsureSchemaIsUpToDate`

martinothamar:
#### Description

Right now, having a solution level config file without a project level one is supported. But `EnsureSchemaIsUpToDate` doesn't care about that and will therefore throw an exception if it doesn't find a config file in the project directory (it only checks for project level, even though earlier in the callstack if neither project or solution level one is found a project level one is generated automatically).
This PR avoids throwing exception during `EnsureSchemaIsUpToDate` when a solution config is specified.

#### The solution

During `EnsureSchemaIsUpToDate`, a `string solutionDirectory` argument has been added, so that both project and solution file can be checked. Will only try to ensure schema is up to date if the respective file exists.

#### Todos

 * [ ] Related issues
 * [x] Tests
 * [ ] Documentation

### DIFF
--- a/Fody/ConfigFileFinder/ConfigFileFinder.cs
+++ b/Fody/ConfigFileFinder/ConfigFileFinder.cs
@@ -99,7 +99,8 @@ public static class ConfigFileFinder
 
     static void CreateSchemaForConfig(string projectConfigFilePath, IEnumerable<WeaverEntry> weavers)
     {
-        var schema = XDocument.Parse(Fody.Properties.Resources.FodyWeavers_SchemaTemplate);
+        var template = Fody.Properties.Resources.FodyWeavers_SchemaTemplate;
+        var schema = XDocument.Parse(template);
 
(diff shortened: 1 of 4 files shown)


Fody/ConfigFileFinder/ConfigFileFinder.cs
```diff
@@ -99,7 +99,8 @@ public static WeaverConfigFile GenerateDefault(string projectDirectory, List<Wea
 
     static void CreateSchemaForConfig(string projectConfigFilePath, IEnumerable<WeaverEntry> weavers)
     {
-        var schema = XDocument.Parse(Fody.Properties.Resources.FodyWeavers_SchemaTemplate);
+        var template = Fody.Properties.Resources.FodyWeavers_SchemaTemplate;
```
tom-englert: Do not introduce unnecessary changes

SimonCropp:
sorry for the delay on this one. i will smoke test it and deploy now

PR #770, opened 2019-11-17
Add Configuration switching for FodyWeavers.xml

arontsang:
Add ability to switch FodyWeavers.xml by Build Configuration.

The user can add a FodyWeavers.Debug.xml file for Debug builds.

### DIFF
--- a/Fody/ConfigFileFinder/ConfigFileFinder.cs
+++ b/Fody/ConfigFileFinder/ConfigFileFinder.cs
@@ -11,22 +11,30 @@ public static class ConfigFileFinder
     static readonly XNamespace schemaNamespace = XNamespace.Get("http://www.w3.org/2001/XMLSchema");
     static readonly XNamespace schemaInstanceNamespace = XNamespace.Get("http://www.w3.org/2001/XMLSchema-instance");
 
-    public static IEnumerable<WeaverConfigFile> FindWeaverConfigFiles(string solutionDirectoryPath, string projectDirectory, ILogger logger)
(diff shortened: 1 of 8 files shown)


arontsang:
This feature most often is requested wrt to Costura.

Turning Costura off for debug makes sense for faster debugging.

tom-englert:
Wouldn't it be more flexible and straight forward to just pass the configuration file name instead of having one hardcoded implicit rule that depends on the existence of a file?

ltrzesniewski:
That would force the user to know how to write MSBuild conditions. It's not very hard though, but something worth considering.

But actually I wonder if it wouldn't be even better to implement a switch in Costura which would disable it in debug builds.

I'm not sure there are many weavers you'd want to configure differently between debug/release builds, and maybe such weavers should implement options of their own (InlineIL does this for instance).

SimonCropp:
does anyone want this for a use case other than costura? given costura is being deprecated...

SimonCropp:
@GeertvanHorrik whats your thoughts on this one?

GeertvanHorrik:
I think this one is only needed for Costura? Maybe method timer, but in that case you can do the #if DEBUG inside the method time logger.

Therefore I think this should be closed since it will make maintenance harder for people to understand. It's a nice idea, but maybe later if we find some real use cases besides costura.

ltrzesniewski:
Thanks for your contribution @arontsang, but this has been superseded by #842, where a more flexible approach has been implemented.

PR #842, opened 2020-04-11
Enable configuring weavers via msbuild

tom-englert:
Since FodyWeavers.xml is not very flexible (see e.g. #770) I would propose to extend the weaver configuration to be controllable by msbuild logic.

This PR is currently just a proof of concept, not final code.

There are no specific tests yet. Temporary verification is the one failing test that shows that overriding works as desired.

Here you can find the sample usage:
https://github.com/Fody/Fody/pull/842/files, see SampleTargetTests.csproj at the end.


tom-englert:
Is it worth finishing this idea?

SimonCropp:
sorry for the delayed reply. yes i think it is worth it. do you envisage us eventually removing the xml file?

tom-englert:
@SimonCropp probably not in the first step, but it could be worth thinking about as a future feature.

ltrzesniewski:
I like the idea, but not particularly the syntax. I'd replace this:

```XML
<WeaverConfiguration Include="SampleWeaver" Attribute="MyProperty" Value="Replacement" />
```

with:

```XML
<WeaverConfiguration Include="SampleWeaver" MyProperty="Replacement" />
```

I'd basically make all metadata items of the `WeaverConfiguration` item available to the weaver. It would make the weavers easier to configure if you want to set several config values.

Also, I suppose some weavers accept child XML elements in their config instead of attributes. That would be harder to replace with MSBuild items.

MSBuild properties [can have arbitrary XML values](https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild-properties?view=vs-2019#store-xml-in-properties) though, so maybe we could store all the weaver config in a MSBuild property? If that works, we would have a nice backwards-compatible way to remove FodyWeavers.xml entirely if we want to.

tom-englert:
Cool, maybe the msbuild-property with xml content is the missing link. I'll have a look.

SimonCropp:
this is a nice idea. i would be good to have the doco for this feature in a PR to Fody/home. i often find writing the doco for a feature can provide some nice feedback into the implementation

> maybe the msbuild-property with xml content is the missing link. I'll have a look

Would need to see how the MSbuild xml looks to make a call on that

tom-englert:
@SimonCropp I'll update this proof of concept to see which strategy works best. If I have a working prototype, I'll start of with the docu.

tom-englert:
It's now according to https://github.com/Fody/Home/pull/187

SampleWeaver.Fody/SampleWeaver.Fody.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFrameworks>netstandard2.0</TargetFrameworks>
```
ltrzesniewski, line 4: That's probably not an intended change.

Issue #851, opened 2020-04-25
Verify correct settings of PrivateAssets


tom-englert:
After installing a weaver via NuGet, the `PrivateAssets="all"` is missing in the `PackageReference` node, which results in the weavers attribute assembly being listed in the copy-local dependencies.

Setting  `<DevelopmentDependency>true</DevelopmentDependency>` in the weavers NuGet package is not an option, since this will create an unusable `<IncludeAssets>` entry, so after installing a weaver the `PrivateAssets="all"` must be set manually. 

Since this is likely to be forgotten or overlooked, Fody should emit a warning/error when a weaver with the `ShouldCleanReference` set to true does not have a corresponding `PrivateAssets="all"` in the project file.

Only usage via `PackageReference` in the project file will be supported.

(Fixes https://github.com/Fody/PropertyChanged/issues/567)

PR #855, opened 2020-05-01
Validate package reference settings

tom-englert:
#851: Verify correct settings of PrivateAssets

### DIFF
--- a/Fody.sln.DotSettings
+++ b/Fody.sln.DotSettings
@@ -3,6 +3,7 @@
 	<s:String x:Key="/Default/CodeInspection/Highlighting/InspectionSeverities/=ArrangeAccessorOwnerBody/@EntryIndexedValue">ERROR</s:String>
 	<s:String x:Key="/Default/CodeInspection/Highlighting/InspectionSeverities/=ArrangeMethodOrOperatorBody/@EntryIndexedValue">ERROR</s:String>
 	<s:String x:Key="/Default/CodeInspection/Highlighting/InspectionSeverities/=ArrangeRedundantParentheses/@EntryIndexedValue">ERROR</s:String>
+	<s:String x:Key="/Default/CodeInspection/Highlighting/InspectionSeverities/=AssignNullToNotNullAttribute/@EntryIndexedValue">DO_NOT_SHOW</s:String>
 	<s:String x:Key="/Default/CodeInspection/Highlighting/InspectionSeverities/=BuiltInTypeReferenceStyle/@EntryIndexedValue">ERROR</s:String>
 	<s:String x:Key="/Default/CodeInspection/Highlighting/InspectionSeverities/=CheckNamespace/@EntryIndexedValue">DO_NOT_SHOW</s:String>
 	<s:String x:Key="/Default/CodeInspection/Highlighting/InspectionSeverities/=ConvertNullableToShortForm/@EntryIndexedValue">ERROR</s:String>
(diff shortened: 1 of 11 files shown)


Fody.sln.DotSettings
```diff
@@ -12,6 +13,7 @@
 	<s:String x:Key="/Default/CodeInspection/Highlighting/InspectionSeverities/=MergeCastWithTypeCheck/@EntryIndexedValue">ERROR</s:String>
 	<s:String x:Key="/Default/CodeInspection/Highlighting/InspectionSeverities/=PartialMethodWithSinglePart/@EntryIndexedValue">ERROR</s:String>
 	<s:String x:Key="/Default/CodeInspection/Highlighting/InspectionSeverities/=PartialTypeWithSinglePart/@EntryIndexedValue">ERROR</s:String>
+	<s:String x:Key="/Default/CodeInspection/Highlighting/InspectionSeverities/=PossibleNullReferenceException/@EntryIndexedValue">DO_NOT_SHOW</s:String>
```
ltrzesniewski, line 16: Are you sure we want to ignore these kinds of errors? Same with the one above.

tom-englert: Yes, because they conflict with nullable ref types - I get contrary warnings from R#

ltrzesniewski: Ok, I see - the R# support isn't foolproof. It should get better in 2020.2 hopefully.

tom-englert:
It was intended to catch all package references in the first shot, even those that might be conditional and not active at the moment. (however conditional package references are not a common use case.)

Having package references in directory.build.props is not a common use case and not supported by Nuget installer, so if you do this manually, you should really know what you are doing.

Passing all package reference information to the inner weaver would have meant a lot of overhead of serializing and deserializing everything, so I went for the simpler solution (KISS/YAGNE).
This will cover everything that the Nuget Installer supports.

ltrzesniewski:
I'd think that using the info provided by MSBuild would be a simpler solution. 😉
If you think serialization is a problem, why not pass the `ShouldCleanReference` value from the inner weaver back to Fody, and then do the check outside of the inner weaver?

ltrzesniewski:
Would you like me to try this alternative implementation?

tom-englert:
@ltrzesniewski if you got some time, give it a try.

ltrzesniewski:
Ok, I'll do that. Actually I think I'll pass the value of the `PrivateAssets` metadata string of the weaver package to the inner weaver, it makes more sense than passing the value of `ShouldCleanReference` back to Fody.

ltrzesniewski:
I added a commit that changes the implementation, but didn't fully test it yet, hence the draft status.
I think you can still review it though @tom-englert.

tom-englert:
@ltrzesniewski the unit tests are not too meaningful with your implementation, since the code is no longer isolated. 
However I manually tested it with the Integration solution, and it behaved as expected, so from my side this is OK.

ltrzesniewski:
Thanks! I haven't had the time to polish the unit tests properly, I merely tried to adapt the test cases you wrote. I'll try to improve on that, I just need to find some time to do that.

tom-englert:
The tests are fine, I think that's all you can do without emulating the whole build process

ltrzesniewski:
I think I should add some tests for the part in `WeavingTask`. That code relies on the fact that `ITaskItem.GetMetadata` returns an empty string (as opposed to null) for a metadata item that doesn't exist. I'd like to capture that in a test, along with some other stuff like making sure the code still works fine when the `PackageReference` doesn't use the same case in the package name as the weaver. I just haven't had the time to write those tests yet. 😕

ltrzesniewski:
Ok, I added the test I wanted.
I rebased the branch because of a merge conflict.

tom-englert:
Looks good, from my POV we can release this. 
@SimonCropp any objections?

PR #897, opened 2020-07-27
Fix Fody#895.

drauch:
Fix Fody#895

### DIFF
--- a/FodyIsolated/InnerWeaver.cs
+++ b/FodyIsolated/InnerWeaver.cs
@@ -96,7 +96,7 @@ public partial class InnerWeaver :
             SplitUpReferences();
             assemblyResolver = new AssemblyResolver(Logger, SplitReferences);
             ReadModule();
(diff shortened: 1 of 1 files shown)


FodyIsolated/InnerWeaver.cs
```diff
@@ -96,7 +96,7 @@ public void Execute()
             SplitUpReferences();
             assemblyResolver = new AssemblyResolver(Logger, SplitReferences);
             ReadModule();
-            if (ModuleDefinition.Types.Any(x => x.Name == "ProcessedByFody"))
+            if (ModuleDefinition.Types.Any(x => x.Name == GetWeavingInfoClassName()))
```
ltrzesniewski, line 99: Could you just introduce a local variable here, so that `GetWeavingInfoClassName()` doesn't get called for each type in the assembly?

drauch: Good point, done.

PR #961, opened 2021-01-17
Keep the original file extension in WeaverTestHelper.ExecuteTestRun

0xced:
Before this pull request, when supplying a custom assembly name, the target file would be renamed with `.dll` when weaving an `.exe` file. This was confusing.


SimonCropp:
@0xced thanks

PR #986, opened 2021-04-01
Fix #985: Build fails when using dotnet cli and strong assembly

### LINKED ISSUE #985: Build fails when using dotnet cli and strong assembly
Hi,

We are using Fody.PropertyChanged in our WPF project and it works perfectly. However, when we try to build the project with **dotnet build** instead of visual studio, we get this error : 
```
MSBUILD : error : Fody: An unhandled exception occurred: [C:\Repo\<>.csproj]
MSBUILD : error : Exception: [C:\Repo\<>.csproj]
MSBUILD : error : Key does not exist. [C:\Repo\<>.csproj]
```

In our csproj, we use these options : 
```
<SignAssembly>true</SignAssembly>
<DelaySign>true</DelaySign>
<AssemblyOriginatorKeyFile><>.snk</AssemblyOriginatorKeyFile>
```

The build succeeds with **dotnet build** if we remove the SignAssembly from the csproj.

Is there something we can do ?

Also, on our build machine, we are able to build the application. However, the application crashes at runtime with this error : 
image

Is it a common issue ?

tom-englert:
Explicitly handle the `<DelaySign>` property.

### DIFF
--- a/Fody/Fody.targets
+++ b/Fody/Fody.targets
@@ -47,6 +47,7 @@
         SolutionDirectory="$(SolutionDir)"
         References="@(ReferencePath)"
         SignAssembly="$(SignAssembly)"
+        DelaySign="$(DelaySign)"
         ReferenceCopyLocalFiles="@(ReferenceCopyLocalPaths)"
         DefineConstants="$(DefineConstants)"
         DocumentationFile="@(DocFileItem->'%(FullPath)')"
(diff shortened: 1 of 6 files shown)


ltrzesniewski:
Looks good by looking at the diff, but I won't be able to review it *"properly"* ATM.

ltrzesniewski (approved):
This fixes the build of the `WithDelaySigning` integration test when using the SDK. :+1:

I'm just wondering if we should add a `dotnet build` step for the integration tests, but it may be done separately.

SimonCropp:
LGTM

SimonCropp (approved):

PR #1008, opened 2021-06-13
Cleanup project files

ltrzesniewski:
This cleans up project files to make them more readable and to remove build warnings.

I also removed `ToolsVersion`/`xmlns` attributes from the project files injected to consumers by NuGet - I hope that's OK as we require MSBuild v16 anyway.


ltrzesniewski:
I'm merging this now to avoid conflicts caused by dependabot, but I'm pinging you @SimonCropp @tom-englert so you're aware of the `ToolsVersion`/`xmlns` changes - we can revert those if you'd prefer.

PR #1015, opened 2021-07-06
Use GitHub Actions for the CI

ltrzesniewski:
This is just a suggestion. Currently we use AppVeyor and Azure Pipelines for the CI. I figured we could move everything to GitHub Actions instead, to consolidate on a single runner and simplify maintenance.

This PR adds a simple workflow just to show what it would look like. Feel free to close it if you prefer the current system.

What do you think @SimonCropp @tom-englert?


SimonCropp:
sounds good to me

tom-englert:
good idea!

ltrzesniewski:
Cool. I'll add a Linux integration job to the workflow (I don't think macOS is necessary if we have Linux), and disable AppVeyor and Azure Pipelines.

We could fix the unit tests on Linux in a later PR, and start using GitHub Actions in the weavers (the workflows will be simpler there).

tom-englert:
@ltrzesniewski how do we now publish the artifacts to nuget.org? With AppVeyor this was just one click.

ltrzesniewski:
Argh, I didn't know it was one click on AppVeyor, so I just uploaded the build artifact manually. I wonder how it works BTW, but I don't have access to Fody on AppVeyor anyway...

Is manual deployment an issue? If it is, I can restore the AppVeyor build, as I'm not sure there's an easy solution like this for GitHub Actions, but I can take a look. We should also consider if we want to migrate the weaver repos to GitHub Actions because of this.

tom-englert:
I just configured my own builds on AppVeyor for Fody and the weavers I contributed to - it's only a few clicks since everything needed is already in the yaml file.

Also the Azure pipelines provide some features neither AppVeyor nor GitHub have, like running the integration tests not only with dotnet but also with  MSBuild and use chained builds to test the output package with various VisualStudio versions.

So let's stay with the AppVeyor and Azure builds in parallel until GitHub can provide all the features.

ltrzesniewski:
Ok I can restore AppVeyor if you prefer.

As for Azure Pipelines, I can add an integration build with MSBuild to GitHub Actions (VS is installed out of the box). That way we could still remove Azure Pipelines. I didn't see Azure Pipelines testing Fody on several VS versions though.

tom-englert:
Azure: Actually it's in two different build definitions (https://dev.azure.com/tom-englert/Fody/_build?definitionId=43)

tom-englert:
Just merged this into one definition: https://github.com/Fody/Fody/pull/1016

ltrzesniewski:
Ok I see what you mean. I suppose we could use the `windows-latest` and `windows-2016` environments on GitHub Actions to achieve the same thing (VS 2019 and VS 2017).

PR #1016, opened 2021-07-08
Add package integration tests

tom-englert:
Add new solution that allows to test the built fody package across different platforms or VS-Versions


ltrzesniewski:
I wonder if we should test VS 2017 at all, given that [only VS 2019 is supported](https://github.com/Fody/Home/blob/master/pages/supported-runtimes-and-ide.md). Also, how does this pass at all, since [Fody has a chack](https://github.com/Fody/Fody/blob/ec2b00e0750eb46589fc33685db83413ad435c83/Fody/Fody.targets#L37-L38) for the VS version? 🤔

PackageIntegrationTest/SampleApplication/SampleApplication.csproj
```diff
@@ -0,0 +1,123 @@
...
+    <PackageReference Include="NullGuard.Fody">
+      <Version>3.0.0</Version>
+      <PrivateAssets>all</PrivateAssets>
+    </PackageReference>
+    <PackageReference Include="PropertyChanged.Fody">
+      <Version>3.4.0</Version>
+      <PrivateAssets>all</PrivateAssets>
+    </PackageReference>
+  </ItemGroup>
+  <Target Name="BypassFodyVSVersionCheck" BeforeTargets="FodyTarget">
```
ltrzesniewski, line 117: Oh, I see, you cheated 🙂

SimonCropp:
yeah i dont see any value in testing against vs2017. mainly because i do t want to be expected to provide support for it

tom-englert:
Yes, VS2017 may look outdated now, but I needed this in 2019 when you kicked us out before we were able to migrate all projects, so I like to have this in place for the upcoming VS2019/2022 migration, so that hassle won't happen again.

When we move the Fody build to VS2022, we just bump this VS2017 build with VS2019, so we can easily verify we don't break anything in the first place.

tom-englert:
Change strategy

Issue #167, opened 2015-04-25
Add support for Incremental Builds


qmfrederik:
The Fody.targets file defines two new targets, FodyTarget and FodyVerifyTaget.
These targets don't have Input and/or Output files defined, so they will run every time.

In larger projects this can increase the build time, as the timestamp of the output assembly can change and this will cause full rebuilds to cascade.

See https://msdn.microsoft.com/en-us/library/ms171483.aspx for more info.

SimonCropp:
sounds good. send us a pull request

SimonCropp:
also note this shortcuts processing https://github.com/Fody/Fody/blob/master/Fody/FileChangedChecker.cs

SimonCropp:
@qmfrederik any more thoughts on this?

qmfrederik:
Hi Simon,

Thanks for the feedback!

It's great to see you shortcut processing; however, the net result is still that MSBuild sees a potential change to the file so it triggers a full rebuild of the solution. 

I'd be happy to send a pull request and would be happy to work on it when I have the time. That may take a while, though.

Quick question: MSBuild provides built-in functionality for detecting whether a task needs to run or not. You specify the input files and output files of the task, and MSBuild looks at the file stamps of the input/output files to determine whether the task needs to be run.

My impression is that for Fody, the input file and output files are the same (the assembly), which complicates things.

Perhaps there's a way to have Fody generate a "phony" file so that the build task can have seperate input/output files specified?

Thanks

SimonCropp:
Shouldnt the inout and the outputs just be the assembly and the pdb?

qmfrederik:
Hi Simon,

You mean Fody changes the assembly and does not make any changes to the .pdb?

Yes, in that case, it may work to set the .pdb to the input file and the assembly to the output file. 

If you rebuild a project, and the code has not changed, I guess the timestamp of the .pdb would be older than the .dll. So in that scenario, Fody would not run again.

I'm not sure what happens on a fresh build. If the timestamp of the .pdb is newer than that of the .dll, we should be OK.

I'll check over the weekend and let you know.

Frederik.

distantcam:
I think this is a non-issue. I tried creating a repo where Fody would keep rerunning on the same code and I couldn't do it.

I think because the fody tasks chain off the `AfterCompile` and `AfterBuild` tasks, if they don't run then the fody tasks don't run.

qmfrederik:
Let me try and build a simple repro and get back to you.

[...]

PR #1112, opened 2022-06-29
Fix incremental builds

pavel1269:
- Was added in 2.0.0 but become broken somewhere in 3.x
  - https://github.com/Fody/Fody/issues/167


pavel1269:
Thank you for the reply.

In a solution we have a project that is always being build. I narrowed it down to `<Import Project="..\packages\Fody.6.6.3\build\Fody.targets" Condition="Exists('..\packages\Fody.6.6.3\build\Fody.targets')" />` causing the build and not "up-to-date".

Going to `Fody.targets` and doing these changes which are outlined in the very old issue i got it to work.

PR #1132, opened 2022-10-02
Fix #1129: Use Mono.Cecil via submodule

### LINKED ISSUE #1129: Update Cecil
Now that https://github.com/jbevain/cecil/issues/856 has been fixed, Cecil can be updated. (see also #1110)

@SimonCropp actually there are only the binaries of Cecil in the Fody repo. 
I assume you just did a local build of https://github.com/Fody/cecil at the fody3 branch, and then manually added the build output to the Fody repo.

Shouldn't we better add Cecil as a submodule to Fody, and build the binaries with Fody?

tom-englert:


SimonCropp (approved):

ltrzesniewski (approved):
I'd have preferred for Cecil to stay in `Lib/Cecil`, but that's just nitpicking. 🙂

Issue #1167, opened 2023-03-19
"The assembly has already been processed by Fody" in nuke


ltrzesniewski:
1. Checkout https://github.com/nuke-build/nuke/tree/feature/adaptive-logging
2. Execute `build.sh` or `build.ps1`
3. There should be an `msbuild.binlog` in the root

Note that calling `build.sh/ps1` will first compile all projects as part of compiling `_build.csproj`. After that, the `_build.csproj` will initiate another compilation. That's just what you end up when you do dogfooding :) Both produced warnings at some point.

The suggested property from #1023 did not change anything, but if it's MAUI specific, that explains it.

_Originally posted by @matkoch in https://github.com/Fody/Fody/issues/895#issuecomment-1475344904_

ltrzesniewski:
I can repro this as I'm getting lots of _"The assembly has already been processed by Fody."_ warnings, but the produced binlog file does not contain those warnings.

matkoch:
At which stage are you getting the warnings? Right at the beginning or rather in the middle?

ltrzesniewski:
a `git clean -xdf` and `build.ps1` shows it's at the second (dotnet run) stage. Looks like the following command produces them:

> C:\Dev\Temp\nuke\.nuke\temp\dotnet-win\dotnet.exe build C:\Dev\Temp\nuke\nuke-common.sln --configuration Debug --no-restore --verbosity Minimal

matkoch:
I checked again:

With `build.ps1` I get them twice. Once from `build.ps1` and once from the Compile target.

With `build.ps1 Clean Compile` I get them only from `build.ps1`. So it has something to do with existing artifacts.

tom-englert:
> Note that calling `build.sh/ps1` will first compile all projects as part of compiling `_build.csproj`. After that, the `_build.csproj` will initiate another compilation. That's just what you end up when you do dogfooding :) Both produced warnings at some point.

Looks like this is the root cause. When you trigger multiple compilations, you should set a property per compilation - like the MAUI build does with `$(_ComputeFilesToPublishForRuntimeIdentifiers)` - so you can distinguish the build stages and switch of unneeded tasks.

ltrzesniewski:
But that would trigger an unnecessary recompile I suppose. 🙁

The weaving task shouldn't be triggered on the second build if csc didn't run. Hooking Fody to `TargetsTriggeredByCompilation` looks like a good solution we could try, but that's not exactly a small change. 😅

[...]

PR #1169, opened 2023-03-23
#1167 assembly has already been processed by fody

tom-englert:
Fixing #1167


ltrzesniewski:
Oh, making this conditional on the MSBuild version is a really good idea! 🙂 

Just a small remark: I think this deserves a minor version update (v6.7)

PR #1224, opened 2023-11-04
Take `$(TreatWarningsAsErrors)` into account

ltrzesniewski:
I figured maybe Fody should take `$(TreatWarningsAsErrors)` into account, and emit errors instead of warnings when this property is set to true.

This PR adds a `FodyTreatWarningsAsErrors` property which is set to `$(TreatWarningsAsErrors)` by default, but can be overridden if needed.

This is a suggestion, tell me what you think. **This could be a breaking change to existing users.**

Note that the same outcome can be achieved without this PR by setting `$(MSBuildTreatWarningsAsErrors)` to true - in that case *any* warning emitted during the build will become an error, not just Fody warnings.

*(sorry for all the edits, I must have pressed a shortcut and the PR got submitted before I finished writing the summary)*

### DIFF
--- a/Fody/Fody.targets
+++ b/Fody/Fody.targets
@@ -12,6 +12,7 @@
     <FodyAfterTargets Condition="($(MsBuildMajorVersion) &lt; 17) AND '$(FodyAfterTargets)'==''">AfterCompile</FodyAfterTargets>
     <TargetsTriggeredByCompilation Condition="'$(FodyAfterTargets)'==''">$(TargetsTriggeredByCompilation);FodyTarget</TargetsTriggeredByCompilation>
     <DisableFody Condition='$(UseWPF)==true AND $(MSBuildProjectName.EndsWith("_wpftmp"))'>true</DisableFody>
+    <FodyTreatWarningsAsErrors Condition="$(FodyTreatWarningsAsErrors) == ''">$(TreatWarningsAsErrors)</FodyTreatWarningsAsErrors>
   </PropertyGroup>
 
   <ItemGroup Condition="Exists($(ProjectWeaverXml))">
@@ -62,6 +63,7 @@
         IntermediateCopyLocalFilesCache="$(IntermediateOutputPath)$(MSBuildProjectFile).Fody.CopyLocal.cache"
         RuntimeCopyLocalFilesCache="$(IntermediateOutputPath)$(MSBuildProjectFile).Fody.RuntimeCopyLocal.cache"
         GenerateXsd="$(FodyGenerateXsd)"
+        TreatWarningsAsErrors="$(FodyTreatWarningsAsErrors)"
       >
 
       <Output
--- a/Fody/BuildLogger.cs
+++ b/Fody/BuildLogger.cs
@@ -7,6 +7,8 @@ public class BuildLogger :
     ILogger
 {
     public IBuildEngine BuildEngine { get; set; } = null!;
+    public bool TreatWarningsAsErrors { get; set; }
+
     string? currentWeaverName;
 
     public virtual void SetCurrentWeaverName(string weaverName) =>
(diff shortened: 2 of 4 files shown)


Fody/Fody.targets
```diff
@@ -12,6 +12,8 @@
     <FodyAfterTargets Condition="($(MsBuildMajorVersion) &lt; 17) AND '$(FodyAfterTargets)'==''">AfterCompile</FodyAfterTargets>
     <TargetsTriggeredByCompilation Condition="'$(FodyAfterTargets)'==''">$(TargetsTriggeredByCompilation);FodyTarget</TargetsTriggeredByCompilation>
     <DisableFody Condition='$(UseWPF)==true AND $(MSBuildProjectName.EndsWith("_wpftmp"))'>true</DisableFody>
+    <FodyTreatWarningsAsErrors Condition="$(FodyTreatWarningsAsErrors) == ''">$(TreatWarningsAsErrors)</FodyTreatWarningsAsErrors>
+    <FodyTreatWarningsAsErrors Condition="$(FodyTreatWarningsAsErrors) == ''">false</FodyTreatWarningsAsErrors>
```
tom-englert, line 16: Isn't it false by default?

tom-englert:
I can't remember seeing warnings from Fody for a long time now, so for me this would be OK.

Fody/Fody.targets
ltrzesniewski: Yes, it is. I could remove that line if you prefer, I added it just in case someone uses Fody with an old or different SDK which doesn't define this property.

ltrzesniewski:
I regularly see the warning from PropertyChanged.Fody which says that there's an `OnFooChanged` method but no `Foo` property. That's the main reason I wrote this.

Fody/Fody.targets
tom-englert: But even if it's not defined at all, it will still default to false

ltrzesniewski: Ok I see, thanks, I removed it.

I thought that MSBuild would raise an error when trying to bind the property to a bool task parameter, but it seems it just ignores it when it's empty.

SimonCropp:
if we take this approach, and someone hits one of these and has msbuild treatwarningsaserrors, can they still suppress fody warnings?

tom-englert:
@SimonCropp yes, you can always set `FodyTreatWarningsAsErrors `to false:
```xml
<FodyTreatWarningsAsErrors Condition="$(FodyTreatWarningsAsErrors) == ''">$(TreatWarningsAsErrors)</FodyTreatWarningsAsErrors>
```

ltrzesniewski:
@SimonCropp not sure which property you're talking about, so to clarify:

- `TreatWarningsAsErrors`: you can override this one with `FodyTreatWarningsAsErrors`
- `MSBuildTreatWarningsAsErrors`: you can't override this one (the PR would have to lower Fody warnings into info messages)

SimonCropp:
in that case i have no objections.

ltrzesniewski:
Fo some reason the integration test fails on mac after the merge. It doesn't look to be related to this PR though. 🤔

ltrzesniewski:
Ok, it worked on the third try ¯\\_(ツ)\_/¯

I noticed this PR only works for warnings emitted by the weavers though. The integration test emits a *"Weavers should reference at least the current major version of Fody"* warning which is not turned into an error. I suppose I should turn warnings emitted by Fody itself into errors as well (and set `FodyTreatWarningsAsErrors` to false for the integration test).

ltrzesniewski:
Regarding my previous comments:

- Integration tests did not declare a dependency to their in-solution weaver, so would sometimes fail to build if that project wasn't built on time. I added the missing dependencies.
- Integration tests were setting `TreatWarningsAsErrors` to `true` too late (after `FodyTreatWarningsAsErrors` was set). I added an explicit line which sets `FodyTreatWarningsAsErrors` to `false` there as we actually expect Fody warnings.
- Warnings emitted by Fody itself are correctly turned into errors when needed.

So, everything should be fine now. I also mentioned `FodyTreatWarningsAsErrors` in the docs.

GeertvanHorrik:
This one bit me since we do write warnings from our weavers when weaving, but they are sometimes "unsolvable". Now the builds are failing, but will using `FodyTreatWarningsAsErrors` to fix back the old behavior.

ltrzesniewski:
Well, sorry about that. On my side, I wanted for people to pay attention to warnings. 😅 

Are you talking about some internal weavers? I hope using `FodyTreatWarningsAsErrors` is ok for you. In any case I'm open for discussion if you need anything else.

GeertvanHorrik:
Yes, just tested, that works great. I think overall this is a great improvement. We can still disable it on the few projects that cause these warnings as errors :)

PR #1269, opened 2024-05-08
Base load contexts on weaver paths

### LINKED ISSUE #1268: `FileLoadException: Assembly with same name is already loaded` when multiple Fody-using projects are referenced

We (Datadog) have sponsored Simon through GitHub - hopefully that's sufficient, if not, just let me know 🙂 

#### Preamble

Where relevant, ensure you are using the current stable versions of the following:

 * Fody - `6.8.0` (latest stable)
 * InlineIL.Fody - `1.8.0` (latest stable) and `1.7.4`
 * MSBuild - `17.8.3+195e7f5a3`
 * .NET Core SDK - (`8.0.100`)

#### Describe the issue

We have been using InlineIL.Fody without issue in the Datadog .NET Client library. We recently had to add it to _another_ project. A third project then references those two:

```
Datadog.Trace.Runner
    |__Datadog.Trace (Uses Fody)
    |__Datadog.Trace.Coverage.Collector (Uses Fody)
```

Once we added Fody to the _Datadog.Trace.Coverage.Collector_ project, the _Datadog.Trace.Runner_ has started failing to build, with the following exception from MSBuild.

```
Fody: An unhandled exception occurred:
Exception: Assembly with same name is already loaded
Type: System.IO.FileLoadException
StackTrace:
   at System.Runtime.Loader.AssemblyLoadContext.InternalLoad(ReadOnlySpan`1 arrAssembly, ReadOnlySpan`1 arrSymbols)
   at System.Runtime.Loader.AssemblyLoadContext.LoadFromStream(Stream assembly, Stream assemblySymbols)
   at IsolatedAssemblyLoadContext.LoadNotLocked(String assemblyPath) in C:\projects\fody\FodyCommon\IsolatedAssemblyLoadContext.cs:line 66
   at InnerWeaver.LoadFromFile(String assemblyPath) in C:\projects\fody\FodyIsolated\AssemblyLoader.cs:line 25
   at InnerWeaver.LoadWeaverAssembly(String assemblyPath) in C:\projects\fody\FodyIsolated\AssemblyLoader.cs:line 18
   at InnerWeaver.InitialiseWeaver(WeaverEntry weaverConfig) in C:\projects\fody\FodyIsolated\InnerWeaver.cs:line 159
   at InnerWeaver.InitialiseWeavers() in C:\projects\fody\FodyIsolated\InnerWeaver.cs:line 147
   at InnerWeaver.Execute() in C:\projects\fody\FodyIsolated\InnerWeaver.cs:line 108
Source: System.Private.CoreLib
TargetSite: System.Reflection.Assembly InternalLoad(System.ReadOnlySpan`1[System.Byte], System.ReadOnlySpan`1[System.Byte])
```

This was breaking our build- ([Public build in Azure Devops](https://dev.azure.com/datadoghq/dd-trace-dotnet/_build/results?buildId=156408&view=logs&j=c333b2e2-3335-5c92-02e4-50f3e033a320&t=1e0afc2c-7b7e-5c03-d953-c426d9532f22)).

I subsequently noticed that we were accidentally referencing two different versions of `InlineIL.Fody`. Fixing both projects to use the same assembly seems to have solved the issue, but it still feels like there may be something to investigate here...

#### Minimal Repro

Unfortunately, this appears to be hard to reproduce as it appears to be a race condition. I have not managed to create a minimal repro yet; I can't reproduce with a simplified version of the setup.

#### Make an effort to fix the bug

For a while I _thought_ that disabling MSBuild node reuse would work ([as it seemed to in the PR](https://github.com/DataDog/dd-trace-dotnet/pull/5470) and [test build](https://dev.azure.com/datadoghq/dd-trace-dotnet/_build/results?buildId=156364&view=results)), but even with `MSBUILDDISABLENODEREUSE=1` we were still seeing failures 🙁 The only thing that fixed it was ensuring that all dependent projects used the same version of `InlineIL.Fody`

ltrzesniewski:
Currently, Fody creates an assembly load context per solution directory. A problem with that approach is that if two projects in the same solution try to load two different versions of the same weaver on a single build node, the second load will fail.

This PR changes this logic to create assembly load contexts based on the *set of weaver assembly paths* which will be used in a given project. Therefore, two projects which use two different versions of the same weaver will get two different load contexts since the weaver path won't be the same (and even if it were the same path, the `WeaversHistory` class already takes care of this case).

Note that I may be missing some context here (I don't know why the load contexts were cached by solution path or why they were cached in the first place).

Closes #1268

### DIFF
--- a/Fody/AssemblyPathSet.cs
+++ b/Fody/AssemblyPathSet.cs
@@ -1,0 +1,33 @@
+using System.Runtime.InteropServices;
+
+namespace Fody;
+
+public sealed class AssemblyPathSet : IEquatable<AssemblyPathSet>
+{
+    readonly HashSet<string> assemblyPaths;
+    readonly int hashCode;
+
+    public IReadOnlyCollection<string> AssemblyPaths => assemblyPaths;
+
+    public AssemblyPathSet(IEnumerable<string> paths)
+    {
+        var stringComparer = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? StringComparer.OrdinalIgnoreCase : StringComparer.Ordinal;
+
+        assemblyPaths = new(paths, stringComparer);
+        hashCode = 0;
+
+        foreach (var path in assemblyPaths.OrderBy(i => i, stringComparer))
+        {
+            hashCode = (hashCode * 397) ^ path.GetHashCode();
+        }
+    }
+
+    public bool Equals(AssemblyPathSet? other) =>
+        other is not null && assemblyPaths.SetEquals(other.assemblyPaths);
+
+    public override bool Equals(object? obj) =>
+        obj is AssemblyPathSet other && Equals(other);
+
+    public override int GetHashCode() =>
+        hashCode;
+}
--- a/Fody/Processor.cs
+++ b/Fody/Processor.cs
@@ -22,8 +22,7 @@ public partial class Processor
     public bool GenerateXsd;
     IInnerWeaver? innerWeaver;
 
-    static Dictionary<string, IsolatedAssemblyLoadContext> solutionAssemblyLoadContexts =
-        new(StringComparer.OrdinalIgnoreCase);
+    static Dictionary<AssemblyPathSet, IsolatedAssemblyLoadContext> loadContexts = new();
 
     public ILogger Logger = null!;
     static readonly object mutex = new();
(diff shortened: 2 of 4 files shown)


tom-englert (approved):

ltrzesniewski:
Just FYI, the 6.8.1 release notes did not include #1224 for some reason, I had to add them manually.

tom-englert:
@ltrzesniewski oops, maybe this was hidden among the hundreds of dependabot PR links that I had to remove.

ltrzesniewski:
Yeah no worries, dependabot is a mess 😅

Issue #1297, opened 2024-09-12
Create subfolders for the test weaver helper


GeertvanHorrik:
**I will create a PR once the idea is approved.**

In Costura, I am in need to separate the unit tests into separate folders. Currently the weaver test helper copies all files into `fodytemp` and renames the assemblies. However, this is not possible when running .net (core) exe files (because it cannot find the dll to run from the exe).

My idea is to create subfolders with the intended name of the assembly, but leave the assembly names themselves intact.

If you feel this might impact too many other projects, I am happy to introduce a new argument / parameter to the helper method to decide the behavior (and default to the current one).

PR #1298, opened 2024-09-12
Add folder name to test weaver

GeertvanHorrik:
See #1297

### DIFF
--- a/FodyHelpers/Testing/WeaverTestHelper.cs
+++ b/FodyHelpers/Testing/WeaverTestHelper.cs
@@ -12,10 +12,10 @@ public static class WeaverTestHelper
         bool runPeVerify = true,
         Action<ModuleDefinition>? afterExecuteCallback = null,
         Action<ModuleDefinition>? beforeExecuteCallback = null,
-        string? folderName = null,
         string? assemblyName = null,
         IEnumerable<string>? ignoreCodes = null,
-        bool writeSymbols = false)
+        bool writeSymbols = false,
+        string? folderName = null)
     {
         assemblyPath = Path.GetFullPath(assemblyPath);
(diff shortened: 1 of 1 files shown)


FodyHelpers/Testing/WeaverTestHelper.cs
```diff
@@ -12,13 +12,20 @@ public static TestResult ExecuteTestRun(
         bool runPeVerify = true,
         Action<ModuleDefinition>? afterExecuteCallback = null,
         Action<ModuleDefinition>? beforeExecuteCallback = null,
+        string? folderName = null,
```
tom-englert, line 15: If you move this new parameter to the end, it should be fully backwards compatible

GeertvanHorrik: I have though about it. You mean behind the assemblyName or really at the end? On the long term, I think it's better to keep them together, but short-term would be best to add at the very end.

ltrzesniewski: I agree the `folderName` parameter should be put at the very end, this avoids a breaking change.

ltrzesniewski (commented):
I had a very similar issue in InlineIL, and I had to basically [rewrite `WeaverTestHelper`](https://github.com/ltrzesniewski/InlineIL.Fody/blob/50ec036db400feeb5a991d9a328d22efffb7e7f5/src/InlineIL.Tests/Weaving/WeaverRunner.cs), so I completely understand the need for this change. 🙂

GeertvanHorrik:
Updated the PR based on feedback. Thanks!

tom-englert (approved):

ltrzesniewski (approved):

PR #1304, opened 2024-11-13
Update cecil to v0.11.6

ltrzesniewski:
This updates cecil to v0.11.6 in order to properly support stuff like `allows ref struct`. I'm not sure it's *really* required, but better stay up to date with the latest changes.

---

This includes other changes since I wanted to make it build properly on CI:

- I had to update Microsoft.Build.Tasks.Core since the old version caused these errors when you have the .NET 9 SDK:

  > error NU1903: Warning As Error: Package 'Microsoft.IO.Redist' 6.0.0 has a known high severity vulnerability, https://github.com/advisories/GHSA-hq7w-xv5x-g34j
  > error NU1903: Warning As Error: Package 'System.Formats.Asn1' 8.0.0 has a known high severity vulnerability, https://github.com/advisories/GHSA-447r-wph3-92pm

  This is caused by the new [transitive reference checks](https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-restore#audit-for-security-vulnerabilities). Another option would be to disable these by setting `<NuGetAuditMode>direct</NuGetAuditMode>`.

- I moved the tests to net9.0 since I was getting this error with the new version of Microsoft.Build.Tasks.Core:

  > Error Message: System.IO.FileNotFoundException : Could not load file or assembly 'Microsoft.Build.Framework, Version=15.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. The system cannot find the file specified.

- I disabled the Windows VS job because VS 17.12 is not yet available on GitHub runners. I ran it locally though.

- There are still issues to fix on the other CI platforms, so I'm posting this as a draft and I'll take a look at these.
 
I'm now wondering if NuGetAuditMode=direct wouldn't have been better, maybe I'll rollback this mess and leave it for later 😅


ltrzesniewski:
It looks like we'll have to wait until all CI runners update their VS version before using net9.0.

I just reverted NuGet audit to direct package references and updated old net6.0 targets to net8.0 in order for everything to build properly.

tom-englert (approved):

SimonCropp (approved):

SimonCropp:
@ltrzesniewski we could use this approach for installing sdks https://github.com/VerifyTests/Verify/blob/main/src/appveyor.yml#L9

but that should not block this PR

ltrzesniewski:
Thanks for the tip! But I'll need to wait for the new VS version to be available anyway, as we're also testing the .NET Framework version of MSBuild. I'll open another PR when it's available. I'll also have to check if updating Microsoft.Build.Tasks.Core won't introduce any breaking change (the major version of the referenced assembly changes apparently).

Does anyone want to include anything else in the next version, or should I just release it? (6.9.0 I suppose)

Issue #1257, opened 2024-02-09
fody is missing NuGet package README file


lyndaidaii:
We've noticed that your package on NuGet.org is missing a README file.

  ## Why READMEs are Important
  A README is essential for users to understand your package's purpose and functionality quickly. It's the first thing users see on NuGet.org and helps them decide if your package meets their needs.

  ## Get Started with READMEs

  ### How to Add a README
  If you're new to NuGet READMEs, follow guidance on [how to add one to your package](https://devblogs.microsoft.com/nuget/add-a-readme-to-your-nuget-package/).

  ### How to write high quality of READMEs
  Follow [a blog post](https://devblogs.microsoft.com/nuget/write-a-high-quality-readme-for-nuget-packages/) with README best practices and a template file to help you get started. 

  ## We Value Your Feedback
  We value your feedback. If you encounter any issues or have suggestions, please reply to this issue.

  Thank you for your contribution to the NuGet community.

ltrzesniewski:
@SimonCropp I'm aware this is an automated post, but I didn't pay attention that the Fody main package didn't have a readme in markdown.

Adding a short one may actually be a good idea, what do you think?

tom-englert:
It has a readme.md, so we just have to include it in the nuget package

ltrzesniewski:
TBH, I don't like when projects just copy/paste their GitHub readme files to the NuGet package. They are too long, and contain non-version specific and/or irrelevant stuff such as the shield badges.

A short readme with a basic description looks *much* better IMO.

ltrzesniewski:
(I'm reopening this issue so we can discuss it)

tom-englert:
I'm open to anything, however if you think there is irrelevant stuff on the readme, why is it then relevant when you read it via GH?

ltrzesniewski:
That's a matter of personal preference I'd say. I think I'm used to seeing long readmes on GitHub, but I prefer the descriptions on NuGet to be short and to the point. I add a *"click here to read more"* type of link to the GitHub page in the NuGet descriptions of my own packages.

Hmm maybe that's it: I expect a *readme* on GitHub, but a *description* on NuGet. 

Stuff like this: [![NuGet Status](https://img.shields.io/nuget/v/Fody.svg)](https://www.nuget.org/packages/Fody/) is relevant on GitHub as it indicates which is the latest released package version, but shouldn't be included in a NuGet description IMO.

ltrzesniewski:
[...]

PR #1305, opened 2024-11-17
Embed a simplified readme in the nuget package

tom-englert:
Proposal for #1257:

image

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,6 +1,6 @@
 <Project>
   <PropertyGroup>
-    <Version>6.9.0</Version>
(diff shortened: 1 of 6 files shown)


ltrzesniewski (approved):

ltrzesniewski:
Oh, wait, I just noticed something: Is the period after "Fody/Home" part of the URL?

tom-englert:
Good point. GitHub renders it correctly, but nuget not :-(

tom-englert:
Yes, I saw this, just was curious what the root cause is and where it fails in the build.

PR #1311, opened 2025-02-13
Avoid constructing `StrongNameKeyPair`.

teo-tsirpanis:

(I tried becoming one but my card was declined)

#### Description

This PR uses an alternative way to pass to Cecil the strong-name key pair, without needing to create legacy `StrongNameKeyPair` objects. This will reduce Fody's reliance on its fork of Mono.Cecil, and unless there are other reasons to use it, I can replace it with the upstream NuGet package in this or a subsequent PR.

#### The solution

Instead of setting a `StrongNameKeyPair` to `WriterParameters.StrongNameKeyPair`, we pass the key file's content directly to the `StrongNameKeyBlob` property. Cecil will read it in a cross-platform way.

Fody has been using the `StrongNameKeyPair.PublicKey` property to set the assembly name's public key, and determine if the file actually points to a private key. This PR stops changing the assembly name because I discovered that [Cecil already does that](https://github.com/jbevain/cecil/blob/5de7d8cbc91f6fd98dcc24b26b8c398db497809c/Mono.Cecil/AssemblyWriter.cs#L106), and manually examines the key blob to determine if it is a private key.

#### Todos

 * [ ] Related issues
 * [ ] Tests
 * [ ] Documentation

### DIFF
--- a/FodyIsolated/ModuleWriter.cs
+++ b/FodyIsolated/ModuleWriter.cs
@@ -9,7 +9,7 @@ public partial class InnerWeaver
 
         var parameters = new WriterParameters
         {
-            StrongNameKeyPair = StrongNameKeyPair,
+            StrongNameKeyBlob = StrongNameKeyBlob,
             WriteSymbols = hasSymbols
         };
 
--- a/FodyIsolated/StrongNameKeyFinder.cs
+++ b/FodyIsolated/StrongNameKeyFinder.cs
@@ -1,12 +1,6 @@
-#if (NETSTANDARD)
-using StrongNameKeyPair = Mono.Cecil.StrongNameKeyPair;
-#else
-using StrongNameKeyPair = System.Reflection.StrongNameKeyPair;
-#endif
-
 public partial class InnerWeaver
 {
-    public StrongNameKeyPair? StrongNameKeyPair;
+    public byte[]? StrongNameKeyBlob;
     public byte[]? PublicKey;
 
     public virtual void FindStrongNameKey()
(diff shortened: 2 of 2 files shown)


ltrzesniewski:
Thanks!

I'll let the others review this since they implemented the current code in the first place 😅 - I'd just like to know if this PR only aims to improve the implementation, or is there anything the old code didn't handle?

I'll add a few links to related stuff here:
- https://github.com/Fody/Fody/pull/523
- https://github.com/jbevain/cecil/pull/513
- https://github.com/Fody/Fody/issues/985
- https://github.com/jbevain/cecil/pull/548

> This will reduce Fody's reliance on its fork of Mono.Cecil, and unless there are other reasons to use it, I can replace it with the upstream NuGet package in this or a subsequent PR.

The crypto stuff is the main reason to have a fork. Besides this, Fody only adds `[Obsolete("Use BaseModuleWeaver.TypeSystem")]` to the `TypeSystem` class and changes Cecil's strong name to match Fody's.

tom-englert:
I already had a quick look, but was not sure what the real benefit of this PR is, and how to test.
I think it just simplifies the code, so I'm fine with this.

BTW, the main reason for the fork is to change the strong name to avoid conflicts when `Mono.Cecil` is already used by the target project.

tom-englert:
BTW shouldn't we then also remove `Mono.Cecil.StrongNameKeyPair` from the fork?

teo-tsirpanis:
> but was not sure what the real benefit of this PR is, and how to test

The benefit of this PR is that we will stop using `StrongNameKeyPair` and remove the need for patching upstream Cecil (jbevain/cecil#803). Even in .NET Framework its use is a code smell, because Cecil [relies on the class' private members](https://github.com/jbevain/cecil/blob/5de7d8cbc91f6fd98dcc24b26b8c398db497809c/Mono.Security.Cryptography/CryptoService.cs#L196-L204).

You can test it by building with `dotnet build` a project that uses Fody and strong-naming. It seems [we have CI coverage](https://github.com/Fody/Fody/blob/7fb73713b1d7bb7a17a21a0549fbe2a13add2a4d/Directory.Build.props#L8)?

> Fody only adds `[Obsolete("Use BaseModuleWeaver.TypeSystem")]` to the `TypeSystem` class

Where is that? I couldn't find it.

> changes Cecil's strong name to match Fody's

Fody can keep building Cecil itself, but change the submodule to point to the upstream repository.

> BTW shouldn't we then also remove `Mono.Cecil.StrongNameKeyPair` from the fork?

Yes.

ltrzesniewski:
> > Fody only adds `[Obsolete("Use BaseModuleWeaver.TypeSystem")]` to the `TypeSystem` class
> 
> Where is that? I couldn't find it.

https://github.com/Fody/cecil/blob/6ac9678a8618de548386f6c6e3dd04306cc9dfe7/Mono.Cecil/TypeSystem.cs#L16

PR #1312, opened 2025-03-15
Refactor packaging and remove `Fody.nuspec`.

teo-tsirpanis:

I am!

#### Description

The `Fody` package is currently being created with a `.nuspec` file which explicitly specifies the locations of every file in the package. This is not ideal because it relies in the exact output paths of the library and its dependencies, and also relies on the solution file to enforce a build order for some projects.

#### The solution

This PR removes `Fody.nuspec` and uses MSBuild facilities to define the structure of the `Fody` package. Some of the techniques used are described [here](https://learn.microsoft.com/en-us/visualstudio/msbuild/tutorial-custom-task-code-generation?view=vs-2019#package-the-task-for-distribution).

Validated by manually inspecting that the files in the `Fody` package built from this branch are the same, with the following exceptions:

* The task assemblies were moved from directories `netclassictask` and `netstandardtask` to `tasks/net472` and `tasks/netstandard2.0` for .NET Framework and .NET Standard respectively.
* ~~Next to the .NET Standard 2.0 task assembly there is a `.deps.json` file. This helps MSBuild resolve the task's dependencies.~~
* ~~`FodyHelpers.xml` was included in the package (I can remove it if I try more).~~

#### Todos

 * [ ] Related issues
 * [ ] Tests
 * [ ] Documentation


FodyCommon/FodyCommon.csproj
```diff
@@ -6,7 +6,7 @@
   </PropertyGroup>
 
   <ItemGroup Condition="$(TargetFramework) == 'netstandard2.0'">
-    <PackageReference Include="System.Runtime.Loader" Version="4.3.0" />
+    <PackageReference Include="System.Runtime.Loader" Version="4.3.0" ExcludeAssets="Runtime" />
```
teo-tsirpanis, line 9: This makes `System.Runtime.Loader.dll` to not get included. It wasn't included before and it wasn't causing any problem.

teo-tsirpanis:
I'm not sure why AzDo fails. 🤔

tom-englert:
> I'm not sure why AzDo fails. 🤔

Seems to be a server problem, after re-starting they all ran successful

Fody/Fody.csproj
```diff
@@ -4,15 +4,22 @@
...
+    <GenerateDependencyFile>true</GenerateDependencyFile>
+    <DisableTransitiveProjectReferences>true</DisableTransitiveProjectReferences>
   </PropertyGroup>
 
   <ItemGroup>
     <ProjectReference Include="..\FodyCommon\FodyCommon.csproj" PrivateAssets="All" />
+    <ProjectReference Include="..\FodyIsolated\FodyIsolated.csproj" PrivateAssets="All" />
     <PackageReference Include="Nullable" Version="1.3.1" PrivateAssets="All" />
-    <PackageReference Include="Microsoft.Build.Tasks.Core" Version="17.11.4" PrivateAssets="All" />
+    <PackageReference Include="Microsoft.Build.Tasks.Core" Version="17.11.4" PrivateAssets="All" ExcludeAssets="Runtime" />
```
tom-englert, line 22: This has already PrivateAssets=All, so ExcludeAssets should be not needed

Directory.Build.targets
```diff
@@ -0,0 +1,16 @@
+<Project>
```
tom-englert, line 1: What is the benefit from putting this into a global scope, and then only enabling for one single project?

Fody/Fody.csproj
```diff
@@ -4,15 +4,22 @@
...
+    <AllowedOutputExtensionsInPackageBuildOutputFolder>.dll;.pdb</AllowedOutputExtensionsInPackageBuildOutputFolder>
+    <FodyPackDependencies>true</FodyPackDependencies>
+    <BuildOutputTargetFolder>tasks</BuildOutputTargetFolder>
+    <GenerateDependencyFile>true</GenerateDependencyFile>
+    <DisableTransitiveProjectReferences>true</DisableTransitiveProjectReferences>
   </PropertyGroup>
 
   <ItemGroup>
     <ProjectReference Include="..\FodyCommon\FodyCommon.csproj" PrivateAssets="All" />
+    <ProjectReference Include="..\FodyIsolated\FodyIsolated.csproj" PrivateAssets="All" />
```
tom-englert, line 20: PrivateAssets on ProjectReference has no meaning

tom-englert (changes requested):

Fody/Fody.csproj
teo-tsirpanis: Without `ExcludeAssets="runtime"` the MSBuild assemblies and several transitive dependencies will be copied to the package, which we don't want.

Directory.Build.targets
teo-tsirpanis: It is enabled for both `Fody` and `FodyHelpers`.

Fody/Fody.csproj
teo-tsirpanis: No, it has. Without it the `Fody` package will depend on a non-existent `FodyIsolated` package.

ltrzesniewski (changes requested):
Here's a comparison of `dotnet pack` between the old and new:

image

That's obviously not good, and these changes seem more complicated than they should be. I'll do a second review after the package contains everything needed.

Fody/Fody.targets
```diff
@@ -2,8 +2,8 @@
   <PropertyGroup>
     <ProjectWeaverXml Condition="$(ProjectWeaverXml) == ''">$(ProjectDir)FodyWeavers.xml</ProjectWeaverXml>
     <FodyPath Condition="$(FodyPath) == ''">$(MSBuildThisFileDirectory)..\</FodyPath>
-    <FodyAssemblyDirectory Condition="$(MSBuildRuntimeType) == 'Core'">$(FodyPath)netstandardtask</FodyAssemblyDirectory>
-    <FodyAssemblyDirectory Condition="$(MSBuildRuntimeType) != 'Core'">$(FodyPath)netclassictask</FodyAssemblyDirectory>
+    <FodyAssemblyDirectory Condition="$(MSBuildRuntimeType) == 'Core'">$(FodyPath)tasks\netstandard2.0</FodyAssemblyDirectory>
+    <FodyAssemblyDirectory Condition="$(MSBuildRuntimeType) != 'Core'">$(FodyPath)tasks\net472</FodyAssemblyDirectory>
```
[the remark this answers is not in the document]
teo-tsirpanis: We use some APIs available only in .NET Framework (mainly related to app domains). And I'm actually planning to change .NET Standard 2.0 to .NET 8 in a subsequent PR (it gives us more APIs and lets us unload the isolated assembly load context).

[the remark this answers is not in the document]
ltrzesniewski: Oh, right. But netstandard2.0 needs to stay for maximum MSBuild compatibility.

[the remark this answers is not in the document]
teo-tsirpanis: We don't have access to the .NET runtime version in the targets file that would let us choose whether to use the `netstandard2.0` or `net8.0` assemblies.

IIUC, builds from Visual Studio use the .NET Framework assemblies either way, so the choice of modern .NET TFM hinges on which .NET SDK versions we want to support. The earliest supported is 8, but I can move it to 6 if you want.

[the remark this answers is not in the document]
tom-englert: @ltrzesniewski that is what the artifact of the CI build looks like:
image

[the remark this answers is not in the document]
ltrzesniewski: Well, yes, but I'd rather have a manual build produce the correct result as well.

Also, does the `.deps.json` file really help?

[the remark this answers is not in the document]
tom-englert: I see, I haven't looked at the local build yet, this is of course a no go

teo-tsirpanis:
@ltrzesniewski the `dotnet pack` behavior seems to be by design (I cleaned-up the repo, ran it and it failed). We set `GeneratePackageOnBuild=true` in `Fody.csproj`, which for some reason means that [the `Build` target does not implicitly run](https://github.com/NuGet/NuGet.Client/blob/27a0506221b083213df0be5c0618299e49c4d066/src/NuGet.Core/NuGet.Build.Tasks.Pack/NuGet.Build.Tasks.Pack.targets#L54-L56).

I've been making the packages on my machine with `dotnet build -c Release` and it works fine.

Fody/Fody.targets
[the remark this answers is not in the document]
teo-tsirpanis: > Also, does the .deps.json file really help?

MSBuild [makes use of it](https://github.com/dotnet/msbuild/blob/main/documentation/specs/task-isolation-and-dependencies.md#task-dependency-resolution) if it exists. I'm not exactly sure what the practical difference is, and can remove it from the package if you want.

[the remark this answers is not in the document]
ltrzesniewski: I see, but I think I prefer the default MSBuild behavior of looking for dependencies in the same directory. It works well.

The deps file is a *long* list of references to things such as `Microsoft.Build.Tasks.Core/17.11.4` with precise version numbers, and I'm afraid it could cause issues we could avoid.

ltrzesniewski:
Hmm, that's linked to https://github.com/NuGet/NuGet.Client/pull/2822 and https://github.com/NuGet/Home/issues/7801 which is supposed to be fixed... I think I'll play with the changes in this PR a bit and try to simplify them.

Do we even want to keep `GeneratePackageOnBuild` BTW?

Fody/Fody.targets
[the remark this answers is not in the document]
teo-tsirpanis: Removed `.deps.json`.

teo-tsirpanis:
With my latest commits, the list of files in `Fody.nupkg` is identical to the package released in NuGet, modulo the different folders of task assemblies of each framework.

Fody/Fody.csproj
tom-englert: https://learn.microsoft.com/en-us/visualstudio/msbuild/common-msbuild-project-items?view=vs-2022#projectreference

teo-tsirpanis: Removed. Turns out the package dependencies were not added either way because the NuGet tooling detected that we bundle the assemblies into the package; that's news to me.

FodyHelpers/FodyHelpers.csproj
```diff
@@ -6,22 +6,19 @@
...
   <ItemGroup>
-    <ProjectReference Include="..\cecil\Mono.Cecil.csproj" ReferenceOutputAssembly="false" />
-    <ProjectReference Include="..\cecil\rocks\Mono.Cecil.Rocks.csproj" ReferenceOutputAssembly="false"/>
-    <ProjectReference Include="..\cecil\symbols\pdb\Mono.Cecil.Pdb.csproj" ReferenceOutputAssembly="false"/>
-  </ItemGroup>
-  <ItemGroup>
-    <Reference Include="Mono.Cecil" HintPath="..\cecil\Lib\$(Configuration)\netstandard2.0\Mono.Cecil.dll" />
-    <Reference Include="Mono.Cecil.Pdb" HintPath="..\cecil\Lib\$(Configuration)\netstandard2.0\Mono.Cecil.Pdb.dll" />
-    <Reference Include="Mono.Cecil.Rocks" HintPath="..\cecil\Lib\$(Configuration)\netstandard2.0\Mono.Cecil.Rocks.dll" />
+    <ProjectReference Include="..\cecil\Mono.Cecil.csproj" PrivateAssets="All" />
```
tom-englert, line 19: Same here with PrivateAssets on ProjectReference - does this have any effect?

tom-englert (changes requested):

FodyHelpers/FodyHelpers.csproj
teo-tsirpanis: It doesn't, I removed it.

Fody/Fody.csproj
teo-tsirpanis: Removed all `PrivateAssets` from `Fody` and `FodyHelpers`, and validated that the nupkgs do not have any dependencies. I would prefer the private asset declarations to remain explicit, if you asked me.

Fody/Fody.csproj
```diff
@@ -32,14 +39,22 @@
...
-  <Target Name="SetPackageDependencies" BeforeTargets="GenerateNuspec">
-    <PropertyGroup>
-      <NuspecProperties>configuration=$(Configuration);version=$(PackageVersion);authors=$(Authors);iconUrl=$(PackageIconUrl);projectUrl=$(PackageProjectUrl);description=$(Description);tags=$(PackageTags)</NuspecProperties>
-    </PropertyGroup>
+  <ItemGroup>
+    <None Include="..\package_readme.md" Pack="true" PackagePath="" />
+    <None Include="Fody.targets" Pack="true" PackagePath="build" />
+  </ItemGroup>
+
+  <!-- ExcludeAssets="compile" doesn't work on ProjectReference, and ReferenceOutputAssembly entirely
```
tom-englert, line 47: That's so much hacking with compiler internals. Wouldn't it be much clearer and easier to understand what's going on  when we use `ReferenceOutputAssembly=false and then just copy the files manually?

tom-englert (changes requested):

Fody/Fody.csproj
tom-englert: e.g. like this:
```xml
  <Target Name="CopyFodyIsolatedToOutput" AfterTargets="CopyFilesToOutputDirectory">
    <ItemGroup>
      <_FodyIsolatedFile Include="..\FodyIsolated\bin\$(Configuration)\$(TargetFramework)\*.*" />
    </ItemGroup>
    <Copy SourceFiles="@(_FodyIsolatedFile)" DestinationFolder="$(OutputPath)" />
  </Target>
```

teo-tsirpanis: My proposed solution does precisely what we want to achieve (remove `FodyIsolated` from the references we pass to the compiler), and I believe it is less fragile and more maintainable down the road.

Manually copying the output of `FodyIsolated` assumes that it gets written to a specific location (which [could change](https://learn.microsoft.com/en-us/dotnet/core/sdk/artifacts-output) in the future), and that the project targets the same frameworks as `Fody`. My proposed solution makes none of these assumptions. Hardcoding output paths also brings back what we were doing in the `.nuspec` file, the removal of which is the objective of this PR.

I will file a feature request to MSBuild to make this scenario easier (a `CopyLocalOutputAssembly` metadata on `ProjectReference`).

tom-englert:
What about moving the NuGet generation to a separate project? This would make all the workarounds obsolete:

```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFrameworks>netstandard2.0;net472</TargetFrameworks>
    <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
    <IncludeBuildOutput>false</IncludeBuildOutput>
    <BuildOutputTargetFolder>tasks</BuildOutputTargetFolder>
    <SuppressDependenciesWhenPacking>true</SuppressDependenciesWhenPacking>
    <PackageId>Fody</PackageId>
  </PropertyGroup>

  <ItemGroup>
    <ProjectReference Include="..\FodyIsolated\FodyIsolated.csproj" />
    <ProjectReference Include="..\Fody\Fody.csproj" />
  </ItemGroup>

</Project>
```

and both `Fody` and `FodyIsolated` must have `<Pack>false</Pack>`

ltrzesniewski:
I thought about it, but didn't have time to try it yet. I have used this approach in a few projects of mine and it works well.

BTW I'd put `<IsPackable>false</IsPackable>` in `Directory.Build.props` - Fody produces lots of useless packages today.

tom-englert:
> Fody produces lots of useless packages today.

Really? Which ones?

ltrzesniewski:
When I run `dotnet pack`, I get this:

image

tom-englert:
I see, I never ran `dotnet pack`, since `build `already does the job

ltrzesniewski:
Oh, and here are examples of how I use a separate project to produce a package:

- In [RazorBlade](https://github.com/ltrzesniewski/RazorBlade/blob/5d7b2898feb3fdf04b27a90e29ff90ae8dcb6279/src/RazorBlade/RazorBlade.csproj#L15-L33)
- In [ZeroLog](https://github.com/Abc-Arbitrage/ZeroLog/blob/390af98d56eef4582fc406317587908f350de68c/src/ZeroLog/ZeroLog.csproj#L21-L32)

tom-englert:
The above sample does exactly what we need in terms of binaries, just metadata and Fody.targets need to be added.

ltrzesniewski:
I haven't tried it yet, but doesn't that `<PackageId>Fody</PackageId>` property interfere with `<ProjectReference Include="..\Fody\Fody.csproj" />`?

tom-englert:
Yes, there is a warning during restore, that project needs to be renamed.

tom-englert:
It's only a POC that this is sufficient to output the binaries as we need them

ltrzesniewski:
I like it if it gets rid of the more complex stuff.

Do you intend to work on it a bit more and push it to this PR?

tom-englert:
@teo-tsirpanis what do you think? Maybe you can do an adaption, since I probably could continue earliest next weekend.

teo-tsirpanis:
I took a first stab at this and created a separate `Fody.Package` project referencing `Fody` and `FodyIsolated`. I got an `error : Ambiguous project name 'Fody'.`, which means that we would have to rename the `Fody` project to something like `Fody.Tasks`, which significantly expands the scope of the PR, and by doing this I feel I will spend more time on this PR than I originally planned.

@tom-englert, feel free to push on my branch whenever you have time, thanks.

---

Also the approach used by RazorBlade and ZeroLog is IMHO an overkill. These are "simple" packages that do not bundle their dependencies and have just an analyzer and some MSBuild files added on top, and the extra packaging project [causes the library to be compiled twice as many times](https://github.com/ltrzesniewski/RazorBlade/blob/5d7b2898feb3fdf04b27a90e29ff90ae8dcb6279/src/RazorBlade/RazorBlade.csproj#L17).

ltrzesniewski:
> Also the approach used by RazorBlade and ZeroLog is IMHO an overkill. These are "simple" packages that do not bundle their dependencies and have just an analyzer and some MSBuild files added on top, and the extra packaging project [causes the library to be compiled twice as many times](https://github.com/ltrzesniewski/RazorBlade/blob/5d7b2898feb3fdf04b27a90e29ff90ae8dcb6279/src/RazorBlade/RazorBlade.csproj#L17).

Yes, I'm aware. Both of these libraries have good reasons to be set up like that, and they're not as simple as they look like at first glance. The multiple compilation is not an issue at all (both of these projects even has some files be compiled 3 times).

Of course, I wouldn't use this approach for more standard libraries though.

tom-englert:
Superseded by #1313

PR #1313, opened 2025-04-08
Refactor packaging and remove `Fody.nuspec` #2.

tom-englert:
Refactor packaging and remove Fody.nuspec - second approach, alternative to #1312


Directory.Build.props
```diff
@@ -7,26 +7,26 @@
     <SignAssembly>true</SignAssembly>
     <AssemblyOriginatorKeyFile>$(MSBuildThisFileDirectory)Key.snk</AssemblyOriginatorKeyFile>
     <NoPackageAnalysis>true</NoPackageAnalysis>
-    <PackageLicenseExpression>MIT</PackageLicenseExpression>
+    <IsPackable>false</IsPackable>
+    <GeneratePackageOnBuild Condition="'$(Configuration)' == 'Release'">true</GeneratePackageOnBuild>
```
ltrzesniewski, line 11: Setting `GeneratePackageOnBuild` to true causes errors when calling `dotnet pack`, see the [note here](https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-pack):

> In some cases, the implicit build cannot be performed. This can occur when `GeneratePackageOnBuild` is set, to avoid a cyclic dependency between build and pack targets. The build can also fail if there is a locked file or other issue. 

I was confused by this, I'd prefer the CLI commands to behave normally.

FodyHelpers/FodyHelpers.csproj
```diff
@@ -1,27 +1,22 @@
...
+﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <TargetFrameworks>netstandard2.0;net472</TargetFrameworks>
-    <GeneratePackageOnBuild Condition="$(Configuration) == 'Release'">true</GeneratePackageOnBuild>
+    <IsPackable>true</IsPackable>
     <GenerateDocumentationFile Condition="$(Configuration) == 'Release'">true</GenerateDocumentationFile>
     <Description>Helpers for Fody.</Description>
     <DebugType>embedded</DebugType>
     <DebugSymbols>true</DebugSymbols>
+    <SuppressDependenciesWhenPacking>true</SuppressDependenciesWhenPacking>
```
ltrzesniewski, line 9: Note that this also causes the dependency groups to be removed from the nuspec:

```xml
    <dependencies>
      <group targetFramework=".NETFramework4.7.2" />
      <group targetFramework=".NETStandard2.0" />
    </dependencies>
````

I don't *think* it's an issue though.

**Review by ltrzesniewski — COMMENTED** [review]
I like it! 👍

File: Directory.Build.props [hunk 3]
```diff
@@ -7,26 +7,26 @@
...
+    <Copyright>Copyright (c) Simon Cropp and contributors</Copyright>
+    <PackageLicenseFile>License.txt</PackageLicenseFile>
```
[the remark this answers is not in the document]
tom-englert: I focused on changing as few things as possible, and it was license file before for the Fody package, but expression for the others - maybe expression is the more modern approach.

Directory.Build.props
tom-englert: Since changing this behavior would require to update all the pipelines, since they all rely on this behavior, I would postpone this to another PR

Directory.Build.props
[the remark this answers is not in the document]
ltrzesniewski: Oh right, I thought there already was an expression specified in the nuspec file, but apparently not.

PR #1317, opened 2025-06-21
Target .NET 8 instead of .NET Standard 2.0 and support unloading the `IsolatedAssemblyLoadContext`.

teo-tsirpanis:

I am.

#### Description

This PR retargets the libraries from .NET Standard 2.0 to .NET 8, allowing us to unload the assemblies inside an `IsolatedAssemblyLoadContext`. This update caused several nullability warnings that were fixed in the same PR.

I picked .NET 8 because as of now it is the earliest supported .NET version. [The docs](https://github.com/Fody/Home/blob/master/pages/supported-runtimes-and-ide.md#visual-studio-support) state that Fody is supported on .NET Core SDK 2.2.105 or higher; we will need to update it (preferably to something like "any currently supported .NET SDK version"). If we want to support earlier .NET SDKs, let me know, we can target .NET 6 or even .NET Core 3.1.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -21,8 +21,7 @@
     <PackageReadmeFile>package_readme.md</PackageReadmeFile>
     <PublishRepositoryUrl>true</PublishRepositoryUrl>
     <Nullable>enable</Nullable>
-    <!-- work around https://github.com/dotnet/msbuild/issues/4303 -->
-    <AddSyntheticProjectReferencesForSolutionDependencies>false</AddSyntheticProjectReferencesForSolutionDependencies>
+    <Nullable Condition="'$(TargetFramework)' != 'net8.0'">annotations</Nullable>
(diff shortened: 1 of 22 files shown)


Directory.Build.props
```diff
@@ -21,8 +21,7 @@
     <PackageReadmeFile>package_readme.md</PackageReadmeFile>
     <PublishRepositoryUrl>true</PublishRepositoryUrl>
     <Nullable>enable</Nullable>
-    <!-- work around https://github.com/dotnet/msbuild/issues/4303 -->
-    <AddSyntheticProjectReferencesForSolutionDependencies>false</AddSyntheticProjectReferencesForSolutionDependencies>
```
teo-tsirpanis, line 25: This change was accidentally moved from #1312 (this PR's branch was based on that PR's branch), but since we no longer use solution dependencies, this property doesn't serve any purpose and we can remove it.

PR #1320, opened 2025-08-19
Update Verify.Xunit and xunit.runner.visualstudio

GeertvanHorrik:
Updates dependencies.


ltrzesniewski (approved):
Sorry, I didn't notice your review request earlier. I don't see why there would be a problem with this kind of change.

